PR #207, opened 2022-10-28
alternate runs view

alexhsamuel:
A variant of the runs view for UAT.  This view shows a limited number of runs, with controls to move to earlier or later runs.


alexhsamuel:
To do:
- merge `RunsView` controls into `RunsList`
- customize other uses of `RunsList`: `RunView` and `JobView`
- reflect control state in `RunsView` URL

PR #226, opened 2022-11-18
compressed output

### LINKED ISSUE #216: compress run output

alexhsamuel:
Adds support for compressing run outputs from script/program jobs.  The compression is performed in the agent when returning.  The output is stored compressed, and is also returned compressed in HTTP responses, if the request indicates support for the compression encoding.  We use Brotli, as this compresses very well and is supported by all modern browsers. 

No schema change to DB, as we already had a compression column in the output table.

Resolves #216


alexhsamuel:
Needs some more tests.

alexhsamuel:
I believe with 7b9ccba, Apsis will detect and degrade gracefully if it encounters a running agent from a previous release that doesn't return compressed output.  So there should be no need to track down and stop all running agents during a release.

PR #282, opened 2023-06-26
Feature/async client poll

alexhsamuel:
Uses async rather than sync HTTP requests for polling agent clients.  Uses the HTTPX client library.

Complications:
- We definitely want to use an HTTP client/session, to get connection pooling and long-lived connections, to minimize the number of TCP connections and TLS negotiations when polling frequently.
- We need to shut down the client and its connections cleanly, to avoid leaking connections and asyncio nagging at shutdown.
- We need to shut down the client _before_ the event loop shuts down, since the shutdown code is async.
- The open fds are tied to a particular event loop, so we can't just use another event loop for shutdown.
- We need to support multiple event loops in one process; for instance, pytest-async does this.

To deal with all of this, we monkeypatch the async event loop to support on-close functions, kind of like atexit for async.  We are careful to use a new HTTP client per event loop, but the same client for the entire life of the loop.  When we create a new event loop, we register its close function to be run when the loop closes.  This synchronizes the lifetime of the client with the lifetime of the loop.


alexhsamuel:
Still need to fix tests.

alexhsamuel:
Not sure how to handle the aiohttp client session shutdown gracefully.
https://github.com/aio-libs/aiohttp/issues/7328

PR #338, opened 2024-07-19
Add Litestream Apsis db replica test

LudovicoRighi:
Add test that mimics the situation where Apsis must be restarted using a db created from a Litestream replica.

### DIFF
--- a/test/int/litestream/test_replica.py
+++ b/test/int/litestream/test_replica.py
@@ -98,7 +98,6 @@ def test_replica():
         assert any(restored_db_name in l for l in log)
 
         # check runs data is accurate in the restored db
-        client = inst.client
         for id, state in zip(run_ids, expected_states):
             assert client.get_run(id)["state"] == state
 
@@ -166,7 +165,6 @@ def test_replica_killing_apsis_and_litestream():
         assert any(restored_db_name in l for l in log)
 
         # check the run is still in the running state after reconnecting to the new Apsis instance
-        client = inst.client
         assert client.get_run(run_id)["state"] == "running"
 
         # check run eventually completes successfully
--- a/test/int/litestream/test_replica.py
+++ b/test/int/litestream/test_replica.py
@@ -15,16 +15,7 @@ JOB_DIR = Path(__file__).parent / "jobs"
 
 
 def is_litestream_available():
-    try:
-        subprocess.run(
-            ["litestream", "version"],
-            check=True,
-            stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
-        )
-    except (subprocess.CalledProcessError, FileNotFoundError):
-        return False
-    return True
+    return True if shutil.which("litestream") else False
 
 
 def start_litestream(db_path, replica_path):
--- a/test/int/litestream/test_replica.py
+++ b/test/int/litestream/test_replica.py
@@ -72,6 +72,7 @@ def test_replica():
             for state in expected_states:
                 run_id = client.schedule("job1", {})["run_id"]
                 inst.wait_run(run_id)
+                assert client.get_run(run_id)["state"] == "success"
                 if state != "success":
                     client.mark(run_id, state)
                 run_ids.append(run_id)
--- a/test/int/litestream/test_replica.py
+++ b/test/int/litestream/test_replica.py
@@ -4,6 +4,7 @@ import subprocess
 import os
 import signal
 from time import sleep
+import pytest
 
 from instance import ApsisService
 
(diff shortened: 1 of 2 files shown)


test/int/litestream/test_replica.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # restart Apsis
+        inst.start_serve()
+        inst.wait_for_serve()
+
+        log = inst.get_log_lines()
+        assert any(restored_db_name in l for l in log)
+
+        # check runs data is accurate in the restored db
+        client = inst.client
```
alexhsamuel: You already have this above.

test/int/litestream/test_replica.py
```diff
@@ -0,0 +1,94 @@
...
+
+from instance import ApsisService
+
+
+JOB_DIR = Path(__file__).parent / "jobs"
+
+# -------------------------------------------------------------------------------
+
+
+def test_replica():
```
alexhsamuel, line 48: Add a [`@pytest.mark.skipif`](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) if litestream isn't available.

test/int/litestream/test_replica.py
```diff
@@ -0,0 +1,94 @@
...
+            inst.start_serve()
+            inst.wait_for_serve()
+
+            # populate apsis db with some data
+            client = inst.client
+            run_ids = []
+            expected_states = ["success", "failure", "error", "skipped"]
+            for state in expected_states:
+                run_id = client.schedule("job1", {})["run_id"]
+                inst.wait_run(run_id)
```
alexhsamuel, line 81: Check that the run succeeded.

alexhsamuel:
LGTM.  This is promising.

Not necessarily here, but it would also be good to test that the replica is usable if litestream goes down hard, i.e. SIGKILL.  Also network partition, which you can't necessarily test, but perhaps suspend the process as a proxy.  Likewise killing Apsis rather than orderly shutdown.

I think stress tests would also be a good idea.  Kill Apsis and/or litestream while Apsis is busy and in particular some runs are running.

LudovicoRighi:
I've added another test where:
- Apsis and Litestream processes are terminated with SIGKILL signal;
- Apsis is killed while there is still a run in the running state;

```
[lrighi]🌐 appl112-ny2 in apsis [ litestream][!?⇡][📦 v0.22.7][🐍 v3.10.12][🅒 prod7][⏱ 3s]
> pytest test/int/litestream/test_replica.py
===================================================================================================================== test session starts =====================================================================================================================
(...)
collected 2 items                                                                                                                                                                                                                                             

test/int/litestream/test_replica.py ..                                                                                                                                                                                                                  [100%]

================================================================================================================ 2 passed in 66.09s (0:01:06) =================================================================================================================
```

> Also network partition, which you can't necessarily test, but perhaps suspend the process as a proxy.

@alexhsamuel, could you please elaborate a bit more on this idea?

alexhsamuel:
Sorry if I wasn't clear.  Please don't add new tests/functionality after the initial PR.  Let's get this passing CI and merged, and use new PRs for further tests.

LudovicoRighi:
Ok sounds good, thanks!

> Let's get this passing CI and merged

@alexhsamuel I don't have permissions for neither of the two, so please try to trigger the GH workflow.

test/int/litestream/test_replica.py
```diff
@@ -13,6 +14,19 @@
...
+    try:
+        subprocess.run(
+            ["litestream", "version"],
+            check=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return False
+    return True
```
alexhsamuel: Use `shutil.which()` for this.

LudovicoRighi: Thanks!

test/int/litestream/test_replica.py
```diff
@@ -15,16 +15,7 @@
...
-        subprocess.run(
-            ["litestream", "version"],
-            check=True,
-            stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
-        )
-    except (subprocess.CalledProcessError, FileNotFoundError):
-        return False
-    return True
+    return True if shutil.which("litestream") else False
```
alexhsamuel:
```suggestion
    return shutil.which("litestream") is not None
```

test/int/litestream/test_replica.py
```diff
@@ -0,0 +1,182 @@
+from pathlib import Path
+from contextlib import closing
+import subprocess
+import os
+import signal
+from time import sleep
+import pytest
```
alexhsamuel:
```suggestion
from   contextlib import closing
import os
from   pathlib import Path
import pytest
import shutil
import signal
import subprocess
from   time import sleep
```

PR #340, opened 2024-07-29
Add Litestream replica stress test

LudovicoRighi:
As suggested here: https://github.com/alexhsamuel/apsis/pull/338#issuecomment-2241202464, I'm adding a stress test.
In particular, in the test 500 runs are started immediately and other 500 are scheduled with a small delay. 

@alexhsamuel let me know if this is a reasonable load to run the test with.

The test takes ~20 sec to pass.

### DIFF
--- a/test/int/litestream/test_replica.py
+++ b/test/int/litestream/test_replica.py
@@ -237,10 +237,8 @@ def test_replica_killing_apsis_and_litestream_with_heavy_load():
             assert all(fut.result()["state"] == "running" for fut in as_completed(runs_states))
             assert all(fut.result()["state"] == "scheduled" for fut in as_completed(sched_runs_states))
 
-            # check runs eventually complete successfully
-            runs_results = [exe.submit(inst.wait_run, r) for r in runs_ids]
-            sched_runs_results = [exe.submit(inst.wait_run, r) for r in sched_runs_ids]
-            assert all(fut.result()["state"] == "success" for fut in as_completed(runs_results))
-            assert all(fut.result()["state"] == "success" for fut in as_completed(sched_runs_results))
+            # check all runs eventually complete successfully
+            all_runs_results = [exe.submit(inst.wait_run, r) for r in runs_ids + sched_runs_ids]
+            assert all(fut.result()["state"] == "success" for fut in as_completed(all_runs_results))
 
         inst.stop_serve()
--- a/test/int/litestream/test_replica.py
+++ b/test/int/litestream/test_replica.py
@@ -1,4 +1,6 @@
+from   concurrent.futures import ThreadPoolExecutor, as_completed
 from   contextlib import closing, contextmanager
+import ora
 import os
 from   pathlib import Path
 import pytest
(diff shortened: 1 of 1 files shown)


test/int/litestream/test_replica.py
```diff
@@ -174,3 +176,69 @@ def test_replica_killing_apsis_and_litestream():
...
+        # start Litestream replica
+        replica_path = inst.tmp_dir / "litestream_replica.db"
+
+        with litestream_replica(inst.db_path, replica_path) as litestream_proc:
+            inst.start_serve()
+            inst.wait_for_serve()
+
+            client = inst.client
+            # populate apsis db with a large number of runs
+            with ThreadPoolExecutor() as exe:
```
alexhsamuel: Have you tried this without the executor?  The API handler is single-threaded async, so it's not clear to me that concurrent access will be significantly faster than sequential.

test/int/litestream/test_replica.py
```diff
@@ -174,3 +176,69 @@ def test_replica_killing_apsis_and_litestream():
...
+                runs_fut = [exe.submit(client.schedule, "sleep", {"duration": "9"}) for _ in range(num_jobs)]
+                sched_runs_fut = [exe.submit(client.schedule, "sleep", {"duration": "1"}, time=ora.now() + 14) for _ in range(num_jobs)]
```
alexhsamuel: Where does the 9 and 14 come from?  Did you determine this empirically while running it yourself?  Seems like the right numbers to use will depend both on `num_jobs` as well as on the environment the test is running in.

test/int/litestream/test_replica.py
LudovicoRighi: I was in fact getting very similar results, that would explain why

test/int/litestream/test_replica.py
LudovicoRighi: > Did you determine this empirically while running it yourself?

Correct.

> Seems like the right numbers to use will depend both on `num_jobs` as well as on the environment the test is running in.

I could easily make these durations dependent on `num_jobs`; as per the environment, which factors would you take into account?

test/int/litestream/test_replica.py
alexhsamuel: OK, so I suggest removing the executor then, to keep things simple.

test/int/litestream/test_replica.py
alexhsamuel: > per the environment, which factors would you take into account?

Nothing useful to suggest.  Just keep in mind that the VM that runs the GH Agent CI might be somewhat slower than your hosts.

Ultimately it would be good to build an integration test setup for Apsis with deterministic time, i.e. the test can roll application time forward explicitly, to avoid these kind of timing issues.

LudovicoRighi: Understood, thank you

test/int/litestream/test_replica.py
LudovicoRighi: Ok sounds good

alexhsamuel:
Great!

PR #341, opened 2024-07-31
Add `rename-job` cmd to `apsisctl`

### LINKED ISSUE #168: tool to rename jobs
Provide a mechanism to rename a job while preserving dependencies (both directions).  Be careful about YAML formatting.

LudovicoRighi:
Fixes https://github.com/alexhsamuel/apsis/issues/168.

### DIFF
--- a/python/apsis/ctl.py
+++ b/python/apsis/ctl.py
@@ -125,6 +125,61 @@ def main():
         help="don't print job changes")
 
     #-------------------------------------------------------------
+    # command: rename-job
+
+    def _replace_occurrences(file_path, old_value, new_value):
+        with open(file_path, 'r') as file:
+            lines = file.readlines()
+
+        modified = any(old_value in line for line in lines)
+
+        if not modified:
+            return
+
+        new_lines = [line.replace(old_value, new_value) for line in lines]
+        with open(file_path, 'w') as file:
+            file.writelines(new_lines)
+
+
+    def cmd_rename_job_and_dependencies(args):
+        old_path = args.old_path.with_suffix(".yaml")
+        new_path = args.new_path.with_suffix(".yaml")
+
+        if not old_path.is_file():
+            raise FileNotFoundError(f"Invalid argument: OLD_JOB_PATH does not exist: '{old_path}'.")
+
+        if new_path.is_file():
+            raise FileExistsError(f"Invalid argument: NEW_JOB_PATH already exists: '{new_path}'.")
+
+        new_path.parent.mkdir(parents=True, exist_ok=True)
+        old_path.rename(new_path)
+
+        old_id = str(old_path.with_suffix(""))
+        new_id = str(new_path.with_suffix(""))
+
+        cur_dir = Path.cwd()
+        if cur_dir.name != "jobs":
+            logging.warning(f"Please, make sure to be at your jobs dir root to find all dependencies: {cur_dir}")
+
+        for root, _, files in os.walk(cur_dir):
+            for file in files:
+                file_path = Path(root) / file
+                if file_path.suffix != '.yaml':
+                    continue
+                _replace_occurrences(file_path, old_id, new_id)
+
+            
+    cmd = parser.add_command(
+        "rename-job", cmd_rename_job_and_dependencies,
+        description="Renames a job and all the dependencies to that job starting from current dir.")
+    cmd.add_argument(
+        "old_path", metavar="OLD_JOB_PATH", type=Path,
+        help="Path of the job to be renamed.")
+    cmd.add_argument(
+        "new_path", metavar="NEW_JOB_PATH", type=Path,
+        help="The new path to use.")
+    
+    #-------------------------------------------------------------
     # command: restart
 
     def cmd_restart(args):


python/apsis/ctl.py
```diff
@@ -124,6 +124,61 @@ def section(name):
...
+    def _replace_occurrences(file_path, old_value, new_value):
+        with open(file_path, 'r') as file:
+            lines = file.readlines()
+
+        modified = any(old_value in line for line in lines)
+
+        if not modified:
+            return
+
+        new_lines = [line.replace(old_value, new_value) for line in lines]
```
alexhsamuel, line 139: This is dangerous.  How do you know that `old_value` won't appear in some other context in the YAML source?

The correct way to do this is to look for the job ID in specific places in the YAML structure where you expect job IDs to occur, and change them there.  This brings a maintenance burden, though; you need to update these locations if the job schema changes.

alexhsamuel, line 139: No need to replace by line; you can `str.replace` the entire contents of the file.

python/apsis/ctl.py
```diff
@@ -124,6 +124,61 @@ def section(name):
...
+        if new_path.is_file():
+            raise FileExistsError(f"Invalid argument: NEW_JOB_PATH already exists: '{new_path}'.")
+
+        new_path.parent.mkdir(parents=True, exist_ok=True)
+        old_path.rename(new_path)
+
+        old_id = str(old_path.with_suffix(""))
+        new_id = str(new_path.with_suffix(""))
+
+        cur_dir = Path.cwd()
```
alexhsamuel, line 160: Take the job directory explicitly, like `check_jobs` does, rather than assuming CWD.

python/apsis/ctl.py
```diff
@@ -124,6 +124,61 @@ def section(name):
...
+        if cur_dir.name != "jobs":
+            logging.warning(f"Please, make sure to be at your jobs dir root to find all dependencies: {cur_dir}")
```
alexhsamuel, line 162: Why warn and then continue with mutations?

python/apsis/ctl.py
```diff
@@ -124,6 +124,61 @@ def section(name):
...
+            raise FileExistsError(f"Invalid argument: NEW_JOB_PATH already exists: '{new_path}'.")
+
+        new_path.parent.mkdir(parents=True, exist_ok=True)
+        old_path.rename(new_path)
+
+        old_id = str(old_path.with_suffix(""))
+        new_id = str(new_path.with_suffix(""))
+
+        cur_dir = Path.cwd()
+        if cur_dir.name != "jobs":
```
alexhsamuel, line 161: This may be your local convention, but nothing in Apsis requires that the jobs dir be named `jobs`.

python/apsis/ctl.py
```diff
@@ -124,6 +124,61 @@ def section(name):
...
+            logging.warning(f"Please, make sure to be at your jobs dir root to find all dependencies: {cur_dir}")
+
+        for root, _, files in os.walk(cur_dir):
```
alexhsamuel, line 164: There already is functionality to enumerate jobs in a jobdir in `apsis.jobs`.

alexhsamuel (changes requested):
This code in its current form is is too hacky to be a `apsisctl` subcommand.

If you want to keep it in this form, please put it somewhere local to you; it doesn't belong in Apsis.  If you want it in Apsis, please implement it correctly, along these lines:
- Create `apsis.jobs.rename_job(jobs_dir, old_name, new_name)` that returns a `JobsDir`.  This should use the YAML structure correctly, not just a textual substitution.
- Add a `dump_jobs_dir(jobs_dir, path)` to write out an entire directory.
- Add an `apsisctl` subcommand that loads the entire jobs dir, renames the job correctly, checks it, and writes out the jobs dir to a new location.

alexhsamuel:
BTW the logic as you implemented it is a shell two-liner:
```
$ mv $oldname.yaml $newname.yaml
$ find . -name \*.yaml -exec sed -i "s#$oldname#$newname#g" {} \;
```

alexhsamuel:
Come to think of it, mutating YAML files is fraught; at the very least, you'll lose comments and whitespace.

I suggest you note the shell two-liner above in your wiki, with an admonition that users run `apsisctl check_jobs` double-check the `git diff` before committing.

LudovicoRighi:
Thank you for reviewing, Alex.

> Come to think of it, mutating YAML files is fraught; at the very least, you'll lose comments and whitespace.

That's why I decided to use textual substitution in the first place:
- for comments, I saw [`ruamel` library](https://pypi.org/project/ruamel.yaml/) could be used to preserve them (even if at the cost of adding a dependency to that library.)
- but for whitespaces, I haven't really found a solution (if not raw text manipulation)

-------------------------------------------------------------------------------------------------------

>  If you want it in Apsis, please implement it correctly, along these lines:

> This should use the YAML structure correctly, not just a textual substitution.

I'm happy to follow you comments for improving my code; however, as you mentioned, there is not really a clean way to deal with yaml files, while preserving both comment and whitespaces. If you think this is a blocker, I'll close this PR and just add the shell two-liner to our wiki.

alexhsamuel:
Yes, I would have started with `ruamel_yaml` as well.  I've used it before only with mixed success.  Whether comments/whitespace are a blocker is sort of up to you.  YAML isn't a great format choice for automatic modification, for these reasons.  Comments seem important to me, whitespace less so.

If we can't find a generally correct solution and you'll go with a semi-manual approach, I think good wiki instructions are better.

PR #419, opened 2024-12-11
Improve `check_job_dependencies_scheduled`

LudovicoRighi:
Hi Alex, I believe this check is a bit too strict as the time window in which to look for dependency runs is very tight.

You are basically assuming that if `job_A` depends on `job_B` the time at which `job_B` is scheduled is earlier (or same) than the time at which `job_A` is scheduled. I believe this assumption is a bit too strong (surely it is for us), also because the difference in times among time zones is not constant during the whole year.

####  `A.yaml` def:
```
schedule:
  type: daily
  tz: Europe/London
  calendar: all
  daytime: '17:00:00'
  
condition:
  - type: dependency
    job_id: B
```

####  `B.yaml` def:
```
schedule:
  type: daily
  tz: Europe/London
  calendar: all
  daytime: '18:00:00'
```
(at any time between 17:00 and 18:00):
```
A: scheduled run A(date=2024-12-12): dependency B(date=2024-12-12) not scheduled in [2024-12-10T17:32:55+00:00, 2024-12-12T17:32:55+00:00)
```

Moreover this also breaks in cases like the following, where the dependency is even scheduled earlier than the dependent.
####  `A.yaml` def:
```
schedule:
  type: daily
  tz: Europe/London
  calendar: all
  daytime: '18:00:00'
  
condition:
  - type: dependency
    job_id: B
    args:
      date: '{{Date(date) - 1}}'
```

####  `B.yaml` def:
```
schedule:
  type: daily
  tz: Europe/London
  calendar: all
  daytime: '17:00:00'
```
```
A: scheduled run A(date=2024-12-11): dependency B(date=2024-12-10) not scheduled in [2024-12-10T17:36:54+00:00, 2024-12-12T17:36:54+00:00)

```
----------------------------------------

Anyways, up to you whether to make the dependency window wider here (i.e., apply this commit https://github.com/alexhsamuel/apsis/commit/cad445621349b2dbcfb0d6ebfc64e0aeea3329bc); as long as we can pass `dep_times` to `check_job_dependencies_scheduled` from our side.
(Note: we have reimplemented `apsisctl check-jobs` to also add some extra checks that are ASD specific, so we directly invoke `check_job_dependencies_scheduled`).

### DIFF
--- a/python/apsis/check.py
+++ b/python/apsis/check.py
@@ -100,7 +100,7 @@ def check_job_dependencies_scheduled(
       none, uses (now, now + 1 day).
     :param dep_times:
       (start, stop) time interval in which to look for scheduled runs of
-      dependencies.  If none, uses (now - 1 day, now + 1 day).
+      dependencies.  If none, uses (now - 2 days, now + 2 days).
     :return:
       Generator of errors indicating apparently unscheduled dependencies.
     """
@@ -115,7 +115,7 @@ def check_job_dependencies_scheduled(
         sched_times if sched_times is not None else (time, time + 86400)
     )
     dep_start, dep_stop = (
-        dep_times if dep_times is not None else (time - 86400, time + 86400)
+        dep_times if dep_times is not None else (time - 86400*2, time + 86400*2)
     )
     dep_ivl = f"[{dep_start}, {dep_stop})"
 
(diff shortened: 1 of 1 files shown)


alexhsamuel:
Ah, sorry, I clearly had intended these intervals to be args to the function, but I never wired up the args.  Here's a fix. <https://github.com/alexhsamuel/apsis/pull/420>

alexhsamuel:
See also #421.  You might as well make `dep_times` very wide, then, perhaps a month.

LudovicoRighi:
Nice, thank you!

PR #443, opened 2025-02-13
Prevent Apsis to freeze when jobs are loading

LudovicoRighi:
This should fix the `woke up late` that we have experienced recently.

Tested on a local instance of Apsis that:
- Apsis startup works fine
- `apsisctl reload-jobs` works fine
- Apsis doesn't freeze anymore while loading jobs

### DIFF
--- a/python/apsis/jobs.py
+++ b/python/apsis/jobs.py
@@ -1,3 +1,5 @@
+import aiofiles
+import asyncio
 import logging
 import os
 from   pathlib import Path
@@ -141,16 +143,6 @@ def jso_to_job(jso, job_id):
     return Job.from_jso(jso, job_id)
 
 
-def load_yaml(file, job_id):
-    jso = yaml.load(file, Loader=yaml.SafeLoader)
-    return jso_to_job(jso, job_id)
-
-
-def load_yaml_file(path, job_id):
-    with open(path) as file:
-        return load_yaml(file, job_id)
-
-
 def dump_yaml(file, job):
     yaml.dump(job_to_jso(job), file)
 
@@ -229,7 +221,7 @@ class JobsDir:
 
 
 
-def load_jobs_dir(path):
+async def load_jobs_dir(path):
     """
     Attempts to loads jobs from a jobs dir.
 
@@ -249,13 +241,31 @@ def load_jobs_dir(path):
 
     jobs = {}
     errors = []
-    for path, job_id in list_yaml_files(jobs_path):
-        log.debug(f"loading: {path}")
-        try:
-            jobs[job_id] = load_yaml_file(path, job_id)
-        except SchemaError as exc:
-            log.debug(f"error: {path}: {exc}", exc_info=True)
-            exc.job_id = job_id
+    semaphore = asyncio.Semaphore(10)
+
+    async def load_job(path, job_id):
+        async with semaphore:
+            log.debug(f"loading: {path}")
+            try:
+                async with aiofiles.open(path, mode='r') as file:
+                    content = await file.read()
+                job_jso = await asyncio.to_thread(yaml.safe_load, content)
+                job = Job.from_jso(job_jso, job_id)
+                return job_id, job, None
+            except SchemaError as exc:
+                log.debug(f"error: {path}: {exc}", exc_info=True)
+                exc.job_id = job_id
+                return job_id, None, exc
+
+    load_tasks = [
+        load_job(path, job_id)
+        for path, job_id in list_yaml_files(jobs_path)
+    ]
+    results = await asyncio.gather(*load_tasks)
+    for job_id, job, exc in results:
+        if job is not None:
+            jobs[job_id] = job
+        if exc is not None:
             errors.append(exc)
 
     jobs_dir = JobsDir(jobs_path, jobs)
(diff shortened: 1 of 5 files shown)


python/apsis/jobs.py
```diff
@@ -1,3 +1,5 @@
+import aiofiles
```
gusostow, line 1: add aiofiles to apsis pip dependencies and conda-recipe

python/apsis/jobs.py
```diff
@@ -249,13 +241,31 @@ def load_jobs_dir(path):
...
-            exc.job_id = job_id
+    semaphore = asyncio.Semaphore(10)
+
+    async def load_job(path, job_id):
+        async with semaphore:
+            log.debug(f"loading: {path}")
+            try:
+                async with aiofiles.open(path, mode='r') as file:
+                    content = await file.read()
+                job_jso = await asyncio.to_thread(yaml.safe_load, content)
```
gusostow: I wouldn't think that actually parsing the yaml is at risk of blocking the event loop but I guess it doesn't hurt.

python/apsis/jobs.py
```diff
@@ -249,13 +241,31 @@ def load_jobs_dir(path):
...
+                job = Job.from_jso(job_jso, job_id)
+                return job_id, job, None
+            except SchemaError as exc:
+                log.debug(f"error: {path}: {exc}", exc_info=True)
+                exc.job_id = job_id
+                return job_id, None, exc
+
+    load_tasks = [
```
gusostow: nit: these aren't tasks yet, they're coroutines. `gather` schedules them as tasks

python/apsis/service/main.py
```diff
@@ -74,7 +74,7 @@ def build_apsis(cfg):
     job_dir = cfg["job_dir"]
     log.info(f"opening jobs dir {job_dir}")
     try:
-        jobs = load_jobs_dir(job_dir)
+        jobs = asyncio.get_event_loop().run_until_complete(load_jobs_dir(job_dir))
```
gusostow, line 77: Use `asyncio.run`. That's now the prefered entrypoint in new versions of Python.

gusostow (approved):

python/apsis/jobs.py
stolyaroleh: Let's not spawn threads every time we attempt to parse few lines of text. I think spawning threads might actually make things both more complicated and slower.
image

python/apsis/jobs.py
```diff
@@ -249,13 +241,31 @@ def load_jobs_dir(path):
...
-        log.debug(f"loading: {path}")
-        try:
-            jobs[job_id] = load_yaml_file(path, job_id)
-        except SchemaError as exc:
-            log.debug(f"error: {path}: {exc}", exc_info=True)
-            exc.job_id = job_id
+    semaphore = asyncio.Semaphore(10)
+
+    async def load_job(path, job_id):
+        async with semaphore:
```
stolyaroleh: Without threads, this semaphore is no longer needed.

stolyaroleh (changes requested):

python/apsis/jobs.py
gusostow: Potentially still based on how aiofiles spawns threads. Would have to check the docs

python/apsis/jobs.py
gusostow: Yeah agreed. It might actually hurt. Blocking the event loop while we parse one yaml file at a time isn't a problem.

python/apsis/jobs.py
LudovicoRighi: The semaphore is still used to limit `aiofiles`.

UPDATE: we double checked again and `aiofiles` would use a reasonable number of threads even without the semaphore; removed.
```
min(32, os.cpu_count() + 4)
```

python/apsis/jobs.py
LudovicoRighi: Nice thank you!

python/apsis/service/main.py
LudovicoRighi: As discussed, I think this would require changing `serve` too, 


By just changing to `asyncio.run()` in fact, we would hit the following issue
```
2025-02-14T15:41:30.251 apsis.jobs               I checking: qa/fut/check fut MF safety configs
2025-02-14T15:41:30.267 apsis.sqlite             D query job_id=None since=None min_timestamp=None → 28 runs
2025-02-14T15:41:30.268 apsis.scheduler          I scheduler starts from 2025-02-14T15:34:07+00:00
  File "/space/asd/conda7/envs/prod7/bin/apsisctl", line 10, in <module>
    sys.exit(main())
  File "/space/lrighi/asd-repos/apsis/python/apsis/ctl.py", line 252, in main
    status = args.cmd(args)
  File "/space/lrighi/asd-repos/apsis/python/apsis/ctl.py", line 185, in cmd_serve
    restart = apsis.service.main.serve(
  File "/space/lrighi/asd-repos/apsis/python/apsis/service/main.py", line 103, in serve
    loop = asyncio.get_event_loop()
  File "/space/asd/conda7/envs/prod7-20250214-005/lib/python3.10/asyncio/events.py", line 656, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'MainThread'.
```

Will take another look to see if we can improve the way the even loop is managed.

stolyaroleh (approved):

python/apsis/service/main.py
gusostow: Sounds good

PR #455, opened 2025-03-06
Use default GC settings

gusostow:
Now that we know more about the problem, I believe we're more likely to shoot ourselves in the foot by running less often.

I should note that we did get one extended GC pause with default settings, if I'm remembering correctly. I'm hoping that moving to overpowered physical machine with less swapping will help us avoid that in the future.

### DIFF
--- a/python/apsis/ctl.py
+++ b/python/apsis/ctl.py
@@ -2,9 +2,6 @@
 Control CLI, for maintenance and running the service.
 """
 
-# Configure GC as early as possible.
-import gc
-gc.set_threshold(100_000, 100, 100)
 import apsis.lib.py
 apsis.lib.py.track_gc_stats(warn_time=0.5)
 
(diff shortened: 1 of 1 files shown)


LudovicoRighi:
Some related failures in the tests
```
python/apsis/service/client.py:155: TimeoutError
----------------------------- Captured stderr call -----------------------------
  File "/opt/hostedtoolcache/Python/3.10.16/x64/bin/apsisctl", line 8, in <module>
    sys.exit(main())
  File "/home/runner/work/apsis/apsis/python/apsis/ctl.py", line 250, in main
    status = args.cmd(args)
  File "/home/runner/work/apsis/apsis/python/apsis/ctl.py", line 171, in cmd_serve
    (gc.enable if bool(cfg_gc.get("enabled", True)) else gc.disable)()
NameError: name 'gc' is not defined
```

gusostow:
Woops

LudovicoRighi (approved):

PR #459, opened 2025-03-14
Reverse runs order

### LINKED ISSUE #357: Load runs from most recent to least recent
Loading all runs in the job page takes some time; it would be nice to load them from the most recent to the least recent one rather than the opposite as it is currently done.

cc @alexhsamuel

LudovicoRighi:
- Fixes https://github.com/apsis-scheduler/apsis/issues/357

Tested with a local instance of Apsis.


python/apsis/runs.py
```diff
@@ -520,7 +520,7 @@ def query(
                 )
             )
 
-        return now(), list(runs)
+        return now(), list(runs)[::-1]
```
gusostow: Descending order is a specific decision for frontend presentation. This change should be in the summary websocket endpoint.

gusostow: Are these even sorted in ascending? I thought they were but I don't see what does that in the code. 

Regardless, probably best to sort explicitly in case an implementation detail changes upstream.

gusostow: Looks like it's depending on that dicts are ordered by insertion order which hasn't always been the case and I don't think can be depended upon.

https://github.com/apsis-scheduler/apsis/blob/ba66eb2435ea1ad63b04e15987f783c695ea3740/python/apsis/runs.py#L351-L353

gusostow (approved):
Approved if my comments are addressed and tests look good.

python/apsis/runs.py
LudovicoRighi: > Descending order is a specific decision for frontend presentation. This change should be in the summary websocket endpoint.

Ok, I can move the change here:

https://github.com/apsis-scheduler/apsis/blob/ba66eb2435ea1ad63b04e15987f783c695ea3740/python/apsis/service/api.py#L506

> Looks like it's depending on that dicts are ordered by insertion order which hasn't always been the case and I don't think can be depended upon.

I think it's actually fine to depend on this feature.

https://docs.python.org/3/whatsnew/3.7.html#summary-release-highlights:~:text=the%20insertion%2Dorder%20preservation%20nature%20of%20dict%20objects%20has%20been%20declared%20to%20be%20an%20official%20part%20of%20the%20Python%20language%20spec.

https://mail.python.org/pipermail/python-dev/2017-December/151283.html

The advantage of doing something like `runs[::-1]` rather than `runs.sort(key=sort_key)` are:
1. the fact that we would not have to keep the `sort_key` in sync with the `timeKey` defined here:
https://github.com/apsis-scheduler/apsis/blob/ba66eb2435ea1ad63b04e15987f783c695ea3740/vue/src/runs.js#L38
2. performance, as reversing is faster than sorting (although I'd expect performances to be not too different)

gusostow: > I think it's actually fine to depend on this feature.

Ah, you're right. Nice.


Yeah I think performance actually is a problem for sorting. We don't want to add ~70ms of CPU bound work per tab.
```
[nav] In [22]: import random
          ...: a = list(range(350_000)); random.shuffle(a)

[ins] In [23]: %%timeit
          ...: sorted(a, reverse=True)
68.9 ms ± 1.54 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

The correct way would be to change the sqlite index so it returns in reverse order but who wants to do that.


Instead, stick with your reversing, but can you add comments in relevant areas that we are depending on runs being queried in ascending order.

LudovicoRighi: > The correct way would be to change the sqlite index so it returns in reverse order but who wants to do that.

Yeah, it doesn't feel worth it now - maybe we could consider it later on if we have a stronger reason.

> Yeah I think performance actually is a problem for sorting. We don't want to add ~70ms of CPU bound work per tab.

Right

> Instead, stick with your reversing, but can you add comments in relevant areas that we are depending on runs being queried in ascending order.

Sure, will do before merging

PR #461, opened 2025-04-16
Make `group_id` mandatory for Procstar jobs

LudovicoRighi:


LudovicoRighi:
Actually, not needed. Opted for more general fix https://git.tudor.com/asd/lib/pull/16329

PR #463, opened 2025-04-25
Make `check_job_dependencies_scheduled` much faster

LudovicoRighi:
This will allow us to run `check_job_dependencies` for an entire year in `asd/jobs` CI.


gusostow (approved):
Am I correct that this is exactly the same caching logic but is just using a hash table instead of linear search through dep instances?

LudovicoRighi:
Correct

PR #468, opened 2025-06-20
Support timeout for Procstar programs

LudovicoRighi:
Tested with tests and with a local instance of Apsis.

-----------

Notice I had to add the `<0.9.0` constraint for `ora`, as the CI build was failing otherwise:
- https://github.com/apsis-scheduler/apsis/actions/runs/15778992429/job/44479825255
```
      building 'ora.ext' extension
      creating build/temp.linux-x86_64-cpython-310/python/ora/ext
      g++ -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Icxx/include -Ipython/ora/ext -I/opt/hostedtoolcache/Python/3.10.18/x64/include/python3.10 -c python/ora/ext/functions.cc -o build/temp.linux-x86_64-cpython-310/python/ora/ext/functions.o -std=c++17 -fdiagnostics-color=always -Wno-dangling-else
      In file included from python/ora/ext/functions.cc:6:
      python/ora/ext/py.hh: In member function ‘ora::py::Long::operator __int128()’:
      python/ora/ext/py.hh:700:42: error: too many arguments to function ‘int _PyLong_AsByteArray(PyLongObject*, unsigned char*, size_t, int, int)’
        700 |   check_not_minus_one(_PyLong_AsByteArray(
            |                       ~~~~~~~~~~~~~~~~~~~^
        701 |     (PyLongObject*) this, (unsigned char*) &val, sizeof(val), 1, 1, 1));
            |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      In file included from /opt/hostedtoolcache/Python/3.10.18/x64/include/python3.10/Python.h:84,
                       from python/ora/ext/py.hh:11:
      /opt/hostedtoolcache/Python/3.10.18/x64/include/python3.10/longobject.h:170:17: note: declared here
        170 | PyAPI_FUNC(int) _PyLong_AsByteArray(PyLongObject* v,
            |                 ^~~~~~~~~~~~~~~~~~~
      python/ora/ext/py.hh: In member function ‘ora::py::Long::operator __int128 unsigned()’:
      python/ora/ext/py.hh:710:42: error: too many arguments to function ‘int _PyLong_AsByteArray(PyLongObject*, unsigned char*, size_t, int, int)’
        710 |   check_not_minus_one(_PyLong_AsByteArray(
            |                       ~~~~~~~~~~~~~~~~~~~^
        711 |     (PyLongObject*) this, (unsigned char*) &val, sizeof(val), 1, 0, 1));
            |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      /opt/hostedtoolcache/Python/3.10.18/x64/include/python3.10/longobject.h:170:17: note: declared here
        170 | PyAPI_FUNC(int) _PyLong_AsByteArray(PyLongObject* v,
            |                 ^~~~~~~~~~~~~~~~~~~
      error: command '/usr/bin/g++' failed with exit code 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for ora
ERROR: Failed to build installable wheels for some pyproject.toml based projects (ora)
Successfully built apsis
Failed to build ora
```


python/apsis/program/procstar/agent.py
```diff
@@ -406,6 +419,7 @@ async def updates(self):
 
             conn_id = self.run_state["conn_id"]
             proc_id = self.run_state["proc_id"]
+            start = ora.Time(self.run_state["start"])
```
gusostow, line 422: Will this break when we roll out the change if run_state in the DB for existing runs doesn't have "start"?

python/apsis/program/procstar/agent.py
```diff
@@ -431,6 +445,19 @@ async def updates(self):
         try:
             tasks = asyn.TaskGroup()
 
+            if self.program.timeout is not None:
+                async def timeout_handler():
+                    await asyncio.sleep(self.program.timeout.duration)
```
gusostow: Don't you want to sleep relative to the start time, not "now" for the case where we're reconnecting to a running process?

python/apsis/program/procstar/agent.py
```diff
@@ -519,15 +546,14 @@ def more_output():
...
-                    f"exit code {exit_code}" if signal is None
-                    else f"killed by {signal}",
-                    meta=meta,
-                    outputs=outputs
+                message = (
+                    f"exit code {res.status.exit_code}" if res.status.signal is None
+                    else f"killed by {res.status.signal}"
                 )
+                if self.timed_out:
+                    elapsed = ora.now() - start
```
gusostow, line 560: It could make debugging the logs confusing if the `elapsed` in the failure message doesn't exactly match the elapsed logged in the timeout task. Maybe either figure out how to make them the same, or don't log the elapsed in the timeout task

python/apsis/program/procstar/agent.py
```diff
@@ -431,6 +445,19 @@ async def updates(self):
...
 
+            if self.program.timeout is not None:
+                async def timeout_handler():
+                    await asyncio.sleep(self.program.timeout.duration)
+                    if not self.stopping and self.proc is not None:
+                        elapsed = ora.now() - start
+                        log.info(f"{self.run_id}: timeout after {elapsed:.0f} s")
+                        self.timed_out = True
+                        timeout_signal = Signals[self.program.timeout.signal]
+                        self.stop_signals.append(timeout_signal)
```
gusostow, line 462: How is `stop_signals` used downstream? I'm not sure this belongs there because this isn't from putting it in the stopping state

test/int/procstar/test_timeout.py
```diff
@@ -0,0 +1,93 @@
...
+        assert res["state"] == "running"
+
+        # Restart Apsis
+        svc.restart()
+
+        # Wait for the run to timeout and fail
+        res = svc.wait_run(run_id, timeout=10)
+        assert res["state"] == "failure"
+        assert res["meta"]["program"]["stop"]["signals"] == ["SIGTERM"]
+        assert res["program"]["timeout"]["duration"] == 2.0
```
gusostow: Can we test that the actual elapsed is close to the 2s timeout, because the current assertion isn't catching the reconnect bug I pointed out above

gusostow (commented):
Mostly concerned about correctness around reconnects, and making this backwards compatible for jobs that are running before this change gets rolled out.

test/int/procstar/test_timeout.py
LudovicoRighi: Thank you, good catch; I thought of reconnect, but missed the main point

python/apsis/program/procstar/agent.py
LudovicoRighi: I think the only usage is this one, so it should be safe considering it's just metadata; if you think it's not the right place though, I could either:
- adding it in a different section of metadata
- or not adding it at all

https://github.com/apsis-scheduler/apsis/blob/843d5fdc7a5176d1bd9d8366a2ba20d0601e3801/python/apsis/program/procstar/agent.py#L509

python/apsis/program/procstar/agent.py
LudovicoRighi: Good point!

python/apsis/program/procstar/agent.py
LudovicoRighi: Will take a look now.

python/apsis/program/procstar/agent.py
LudovicoRighi: Right

> or don't log the elapsed in the timeout task

I'll do this

LudovicoRighi (commented):
> and making this backwards compatible for jobs that are running before this change gets rolled out.

I haven't thought of backward compatibility, will take a look now.

Otherwise I guess I could pick a good moment tomorrow when only few jobs are running and kill them before restarting Apsis.

python/apsis/program/procstar/agent.py
gusostow: Let's not add at all

gusostow (approved):
Looks good, assuming you think you can roll out despite backwards compatibility problem with reconnected runs.

PR #471, opened 2025-06-30
Don't reconnect summary websocket 

gusostow:
The idea is that if the run summary websocket is closed we shouldn't reconnect to avoid the doom loop of fetching the full history. This will display an error telling the user to reload the page.


LudovicoRighi (approved):

vue/src/views/RunView.vue
```diff
@@ -287,6 +289,8 @@ export default {
         },
         () => {},
         err => { console.log(err) },
+        () => {},
+        { },
```
LudovicoRighi, line 293: This means we'll keep the existing behavior (i.e. reconnecting) for this socket, right? Could a very large output make Apsis to end up in the same doom loop?

gusostow: Potentially. I wanted to reduce the radius of this change still though because we've never observed an issue on the other websocket endpoints. 

If it ever could happen it would be much rarer for people to be on run page with hundreds of megabytes of output. Also the scheduler doesn't need to json serialize or even compress the logs, so it wouldn't even push CPU up.

PR #473, opened 2025-07-03
Bugfix: transition runs to error when final FdData isn't received

gusostow:
Fixes a bug where runs will hang forever in the running state, waiting for final FdData, which could be dropped due to network interruption, or if it otherwise can't be sent.

This change transitions the run to an error after a 30s timeout. The message includes the real proc exit code so users understand that the run completed, and whether it was successful.

I decided to transition into an error because we're still observing this issue in the wild, so it's good to be more aware of it. Also because the existing output/fddata interfaces put up a sufficient amount of friction for appending an apsis-generate message into the run logs.

### DIFF
--- a/python/apsis/program/procstar/agent.py
+++ b/python/apsis/program/procstar/agent.py
@@ -26,7 +26,7 @@ ntemplate_expand = or_none(template_expand)
 
 SUDO_ARGV_DEFAULT = ["/usr/bin/sudo", "--preserve-env", "--set-home"]
 
-FD_DATA_TIMEOUT = 30.0
+FD_DATA_TIMEOUT = 60
 
 if_not_none = lambda k, v: {} if v is None else {k: v}
 
--- a/python/apsis/program/procstar/agent.py
+++ b/python/apsis/program/procstar/agent.py
@@ -321,12 +321,12 @@ class BoundProcstarProgram(base.Program):
 #-------------------------------------------------------------------------------
 
 async def collect_final_fd_data(
-    intitial_fd_data: FdData | None, proc: Process, expected_length: int
+    initial_fd_data: FdData | None, proc: Process, expected_length: int
 ) -> FdData | None:
     async for update in proc.updates:
         match update:
             case FdData():
-                fd_data_local = _combine_fd_data(intitial_fd_data, update)
+                fd_data_local = _combine_fd_data(initial_fd_data, update)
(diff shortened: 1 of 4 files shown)


python/apsis/program/procstar/agent.py
```diff
@@ -318,6 +320,26 @@ def connect(self, run_id, run_state, cfg):
 
 #-------------------------------------------------------------------------------
 
+async def collect_final_fd_data(
+    intitial_fd_data: FdData | None, proc: Process, expected_length: int
```
LudovicoRighi:
```suggestion
    initial_fd_data: FdData | None, proc: Process, expected_length: int
```

Nit: typo

python/apsis/program/procstar/agent.py
```diff
@@ -318,6 +320,26 @@ def connect(self, run_id, run_state, cfg):
...
+) -> FdData | None:
+    async for update in proc.updates:
+        match update:
+            case FdData():
+                fd_data_local = _combine_fd_data(intitial_fd_data, update)
```
LudovicoRighi:
```suggestion
                fd_data_local = _combine_fd_data(initial_fd_data, update)
```

python/apsis/program/procstar/agent.py
```diff
@@ -318,6 +320,26 @@ def connect(self, run_id, run_state, cfg):
...
+                # specified in the result.
+                assert fd_data_local.interval.start == 0
+                assert fd_data_local.interval.stop == expected_length
+                return fd_data_local
+
+            case _:
+                log.debug("expected final FdData")
+
+    # If we get here, the async iterator ended without FdData
+    return intitial_fd_data
```
LudovicoRighi:
```suggestion
    return initial_fd_data
```

test/unit/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+        pass
+
+    async def request_fd_data(self, *args, **kwargs):
+        pass
+
+    async def send_signal(self, *args, **kwargs):
+        pass
+
+    async def send_signal(self, *args, **kwargs):
+        pass
```
LudovicoRighi:
```suggestion
    async def send_signal(self, *args, **kwargs):
        pass
```

Nit: duplicated

LudovicoRighi (approved):

PR #485, opened 2025-07-16
Mitigate websocket disconnects

gusostow:
The browser isn't reconnecting the summary socket after https://github.com/apsis-scheduler/apsis/pull/471. Disconnects are happening a too often.

The underlying issue is a ping/pong timeout which can't complete when the browser can't keep up with summary messages, closing the TCP receive window. 

I'm hoping that increasing the ping timeouts from 20s -> 300s will give the browser a chance to load the full summary history before getting timed out. Then it will only have to handle incremental updates which are much lighter.

### DIFF
--- a/python/apsis/service/api.py
+++ b/python/apsis/service/api.py
@@ -4,6 +4,7 @@ import logging
 import ora
 import re
 import sanic
+import time
 import ujson
 from   urllib.parse import unquote, parse_qs
 import websockets
@@ -504,6 +505,8 @@ SUMMARY_MSG_TYPES = {
 
 @API.websocket("/summary")
 async def websocket_summary(request, ws):
+    connect_time = time.monotonic()
+
     # request.query_args doesn't work correctly for ws endpoints?
     init = "init" in request.query_string.split("&")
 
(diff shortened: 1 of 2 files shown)


python/apsis/service/api.py
```diff
@@ -551,6 +554,14 @@ async def websocket_summary(request, ws):
         except websockets.ConnectionClosed:
             log.debug(f"{prefix} closed")
 
+        except asyncio.CancelledError:
+            # task is cancelled in the event of abnormal websocket closure, usually if
+            # websocket close due to a ping timeout can't complete successfull
```
LudovicoRighi:
```suggestion
            # websocket close due to a ping timeout can't complete successfully
```

python/apsis/service/main.py
```diff
@@ -50,6 +50,10 @@ def get(self, path, method, host):
 app = sanic.Sanic("apsis", router=Router(), log_config=SANIC_LOG_CONFIG)
 app.config.LOGO = None
 
+# more forgiving for when slow browsers can't keep up with websocket messages
+app.config.WEBSOCKET_PING_TIMEOUT = 300
+app.config.WEBSOCKET_PING_INTERVAL = 100
```
LudovicoRighi, line 55: I'm slightly concerned that increasing these values so much (i.e. default values are `20` for both) could have some side effects; for instance, this will cause dead connections to be kept around for longer, right? Do you think this could cause any issues, i.e. increasing the load on Apsis?

gusostow: My understanding is that the only case where the connection would actually be closed without apsis realizing is if there's a true network interruption, which should be rare in our system and we'd also have bigger problems than a 5min wait.

Setting it low is good for detecting back pressure and closing the connection, but that's exactly what we don't want.

LudovicoRighi (approved):

PR #488, opened 2025-07-16
Make `SCHEDULE` button more obviously disabled

LudovicoRighi:


aodinaev:
agreed with the intention but have no clue if this change technically correct or not.

gusostow (approved):
Awesome

gusostow:
> agreed with the intention but have no clue if this change technically correct or not.

Same, but trusting you tested that this works as intended

PR #500, opened 2025-07-31
Add global runtime timeout default and make it configurable

### LINKED ISSUE #449: Support global timeout for `running` state
We used to have `TIMEOUT` set to `36h` (see https://github.com/alexhsamuel/apsis/commit/7ec3ad94866a925ec0f0cdb819b02c1b31c2e1b4) which was limiting the maximum duration of a run in the running state.

We may want to introduce something like this again; right now, there is no limit at all:
- http://apsis.asd:5000/runs/r7435362

cc @gusostow

LudovicoRighi:
- Fixes: https://github.com/apsis-scheduler/apsis/issues/449

Adding a global configurable runtime timeout value.

--------

I've added to this PR also a fix to the config validation; happy to move this to a different PR if preferred.
In fact, notice that after the first commit the tests were failing with this error:
```
2025-08-01T11:10:00.774 apsis.run_log            E run r8: error: internal: '<' not supported between instances of 'NoneType' and 'float'
  File "/home/runner/work/apsis/apsis/python/apsis/running.py", line 106, in _process_updates
    update = await anext(updates)
  File "/home/runner/work/apsis/apsis/python/apsis/program/agent.py", line 307, in updates
    if self.program.timeout.duration < elapsed:
TypeError: '<' not supported between instances of 'NoneType' and 'float'
```

I have then fixed the issue with this commit https://github.com/apsis-scheduler/apsis/pull/500/commits/f3c35be69a637ebf1b235e5b2c0f8ab1cea63cfd by ensuring we don't set to `None` values that are actually missing in the config due to a bug in `_check_duration`.

### DIFF
--- a/python/apsis/program/base.py
+++ b/python/apsis/program/base.py
@@ -158,7 +158,7 @@ class Timeout:
 
 def get_global_runtime_timeout(cfg):
     timeout_duration = get_cfg(cfg, "program.timeout.duration", None)
-    if timeout_duration is None or timeout_duration <= 0:
+    if timeout_duration is None:
         return None
     timeout_signal = get_cfg(cfg, "program.timeout.signal", TIMEOUT_SIGNAL)
     return Timeout(duration=timeout_duration, signal=timeout_signal)
--- a/python/apsis/program/base.py
+++ b/python/apsis/program/base.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from signal import Signals
 
 from apsis.lib import memo
 from apsis.lib.api import decompress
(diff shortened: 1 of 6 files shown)


python/apsis/program/base.py
```diff
@@ -153,6 +156,14 @@ def bind(self, args):
         return type(self)(duration=duration, signal=signal)
 
 
+def get_global_runtime_timeout(cfg):
+    timeout_duration = get_cfg(cfg, "program.timeout.duration", None)
+    if timeout_duration is None or timeout_duration <= 0:
```
gusostow: Can't we trust that `timeout_duration` is valid after config parsing assertions?

gusostow (approved):

python/apsis/program/base.py
LudovicoRighi: Ah yeah right! Forgot to update that

PR #511, opened 2026-01-15
Modify release workflow to keep `uv.lock` up to date

LudovicoRighi:
Automating this step:
- https://github.com/apsis-scheduler/apsis/pull/509


gusostow (commented):
We shouldn't lock as part of a release workflow because it re-resolves dependencies and could break tests.

Instead, on PR tests we should install the test environment with `uv sync --locked` or similar to fail if the lockfile isn't up-to-date.

PR #514, opened 2026-01-20
Remove old non-Procstar agents

LudovicoRighi:
> Clean up the classic agents.

This PR is just to complete this ^ TODO that Alex left.

Procstar is much better than old classic agents; better to remove them, also to keep the repo clean and tidy.

### Testing

I've tested this locally using:
- a copy of our prod `apsis.db`
- a copy of our prod jobs (modified to just run a simple print command)

I've also verified that all our dependencies from `lib` to Apsis still work fine.


docs/programs.rst
```diff
@@ -12,8 +12,6 @@ configuration of each program type.
 Apsis provides these program types, and you can extend Apsis with your own.
 
 - `no-op`: Does nothing.
-- `shell`: Executes a shell command, possibly on another host.
-- `program`: Invokes an executable directly, with command line arguments.
```
gusostow, line 16: Are you against leaving the local process programs. Seems like they still could be useful for testing and demonstration purposes?

Don't feel strongly though.

gusostow (approved):
Ok to remove process programs if you think it will simplify things.

docs/programs.rst
LudovicoRighi: Yeah, I'd prefer to remove all the code we don't need anymore.

> Seems like they still could be useful for testing and demonstration purposes?

For these purposes we're already using Procstar, which is better for more proper testing.

Issue #333, opened 2024-04-18
Procstar ECS program


gusostow:
The plan we discussed a while ago: 

The apsis scheduler launches jobs as ECS tasks, which are basically just containers. The running process in the container is still mediated by procstar: the container CLI command contains the json spec, then procstar and the ECS task terminate once the job process exits. 

We said, the scheduler would poll and signal the job via the ECS API, rather than using procstar itself.  But now I'm wondering if it's reasonable to lean more on procstar for tracking and controlling jobs instead of ECS. 

- ECS requires polling which we want to move away from.
- Easier and more unified with non ECS procstar programs.
- Could support incrementally returning log outputs in the future, which it looks like you've laid the groundwork for. 

Is it possible to launch a one-off procstar process then connect to it via the apsis scheduler's websocket server? I'm not seeing a straightforward way to retrieve the connection for the new procstar agent as-is.

alexhsamuel:
I don't remember the plan involving polling via ECS.  I think using the procstar agent's websocket communication mechanism is preferable, as long as the ECS instance routes back to Apsis.

Could we start by discussing requirements?
- Do you necessarily need a new procstar instance on a new ECS instance per run?  What is the motivation?  This will cost substantial startup time.
- Do you want Apsis to start the ECS instance?  If the ECS instance outlives a single run, do you want Apsis to manage an elastic pool of ECS + procstar instances, starting/stopping them based on some criteria?
- Would you want to run multiple runs on a single ECS + procstar instance?  This probably is more economical than one-run-per-ECS.

The procstar agent connects to Apsis (not vice versa), which is why routing has to be enabled.  There is no need to have multiple procstar agents on one ECS instance (assuming single user) since one agent can handle multiple processes.

One way to associate a particular Apsis run with a particular procstar agent is via the group ID; you can make up as many of these as you want, and they may be ephemeral. But I'm not sure this is a good idea&mdash;you probably want some abstraction between your job configs and specific ECS instances.  A job probably wants to specify a username (or some other auth signifier) and possibly some resource info like ECS instance type, and leave it to Apsis to arrange for a new or existing ECS instance to run on, right?

gusostow:
I think might've misremembered a decision to poll with ECS, especially because it's so inferior to procstar websocket updates.

[...]

PR #516, opened 2026-01-29
Support running Procstar jobs on AWS ECS

LudovicoRighi:
- See https://github.com/apsis-scheduler/apsis/issues/333

Adding support for running Apsis run on Procstar agents launched on AWS ECS as ECS Tasks.

### TODO:
- Before merging this PR, make sure we can safely rollback even if ECS runs are in the `apsis.db` 
- Once this PR is approved, consider whether we want to roll it out as-is, or split into refactor (non-ECS) and ECS feature addition.
  - (assuming we have good rollback procedure, I believe it should be fine to roll out as a single PR, as I've tried very hard to minimize as much as possible change to non-ECS Procstar agents; to be discussed together)

---------------

UPDATE: 
- merge this PR only after https://github.com/apsis-scheduler/apsis/pull/518 (see detailed plan there)


docs/config.rst
```diff
@@ -294,3 +294,94 @@ updates for a run from the agent running it.  If null or omitted, Apsis does not
...
+
+    procstar:
+      agent:
+        ecs:
+          cluster_name: my-ecs-cluster
+          task_definition: procstar-agent-task
+          container_name: procstar-agent
+          capacity_provider: procstar
+          region: us-east-1
+          aws_account_number: "123456789012"
```
gusostow: I believe `aws_account_id` is more standard

docs/config.rst
```diff
@@ -294,3 +294,94 @@ updates for a run from the agent running it.  If null or omitted, Apsis does not
...
+
+.. code:: yaml
+
+    procstar:
+      agent:
+        ecs:
+          cluster_name: my-ecs-cluster
+          task_definition: procstar-agent-task
+          container_name: procstar-agent
+          capacity_provider: procstar
```
gusostow: I'm surprised capacity_provider needs to be a public setting. I thought that just needed to be used by ECS cluster on the backend?

docs/programs.rst
```diff
@@ -131,6 +132,156 @@ In this example, Apsis sends SIGTERM to the program after five minutes, if it
 hasn't completed yet.  The `signal` key is optional and defaults to SIGTERM.
 
 
+Procstar ECS Programs
+---------------------
+
+Apsis can run programs on AWS ECS (Elastic Container Service) using Procstar
+agents running inside ECS tasks. Unlike standard Procstar programs that connect
+to pre-existing agents, ECS programs launch a new ECS task for each run,
```
gusostow: Worth clarifying that the ECS task is a one-off agent tied to the lifetime of a single run.

docs/programs.rst
```diff
@@ -131,6 +132,156 @@ In this example, Apsis sends SIGTERM to the program after five minutes, if it
...
+
+Exactly one of ``command`` or ``argv`` must be specified.
+
+
+Resource Configuration
+^^^^^^^^^^^^^^^^^^^^^^
+
+ECS programs support the following resource parameters:
+
+- ``mem_gb``: Memory allocation in GiB for the container (e.g., 2 for 2 GB)
```
gusostow: nit: GiB (gibibyte) and GB (gigabyte) are different units

docs/programs.rst
```diff
@@ -131,6 +132,156 @@ In this example, Apsis sends SIGTERM to the program after five minutes, if it
...
+- ``vcpu``: vCPU allocation for the container (e.g., 1 for 1 vCPU)
+- ``disk_gb``: Ephemeral EBS storage in GiB (default configured globally)
```
gusostow: Aren't the default configured globally for mem_gb/vcpu also?

docs/programs.rst
```diff
@@ -131,6 +132,156 @@ In this example, Apsis sends SIGTERM to the program after five minutes, if it
...
+            signal: SIGTERM
+            grace_period: 30s
+
+
+ECS Task Definition Requirements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ECS task definition used by Apsis must meet these requirements:
+
+1. **Container name**: The container running the Procstar agent must have a name
```
gusostow, line 274: Why does apsis care what the container name is?

python/apsis/program/procstar/aws/ecs.py
```diff
@@ -0,0 +1,153 @@
...
+                "name": "procstar-data",
+                "managedEBSVolume": {
+                    "sizeInGiB": disk_gb,
+                    "volumeType": "gp3",
+                    "iops": 3000,
+                    "throughput": 125,
+                    "encrypted": True,
+                    "filesystemType": "ext4",
+                    "roleArn": self.ebs_volume_role_arn,
+                    # TODO: check if we want to retain volumes instead
```
gusostow: deletion is fine. This is just scratch space

python/apsis/program/procstar/aws/ecs.py
```diff
@@ -0,0 +1,153 @@
...
+        return self._ecs_client
+
+    def _create_ebs_volume_config(self, disk_gb: int) -> List[Dict]:
+        return [
+            {
+                "name": "procstar-data",
+                "managedEBSVolume": {
+                    "sizeInGiB": disk_gb,
+                    "volumeType": "gp3",
+                    "iops": 3000,
```
gusostow: Low iops and throughput can bottleneck S3 downloads. We should use the max of 16K iops and 1000 throughput if possible, but it might only be available for volumes of a certain size.

You should figure out the requirement and make them as large as possible by clipping them to the max for a given disk size.

docs/programs.rst
gusostow: Oh it's so you can override its values. I'd just remove this as a config parameter and force the user to use a standard name. But I don't feel strongly.

python/apsis/program/procstar/aws/ecs.py
```diff
@@ -0,0 +1,153 @@
...
+            overrides["taskRoleArn"] = task_role_arn
+            logger.info(f"Overriding task role to: {task_role_arn}")
+
+        volume_configurations = self._create_ebs_volume_config(actual_disk_gb)
+
+        run_task_params = {
+            "cluster": self.cluster_name,
+            "taskDefinition": task_definition,
+            "volumeConfigurations": volume_configurations,
+            "capacityProviderStrategy": [
```
gusostow: Can you set a default capacity provider strategy on the cluster? It seems like this isn't configurable per-job anyways

python/apsis/program/procstar/aws/ecs.py
```diff
@@ -0,0 +1,153 @@
...
+                    "base": 0,
+                }
+            ],
+            "overrides": overrides,
+        }
+
+        if tags:
+            run_task_params["tags"] = tags
+
+        response = await asyncio.get_event_loop().run_in_executor(
```
gusostow, line 139: Is there a reason why you decided not to use aiobotocore? No strong feelings, just curious

python/apsis/program/procstar/aws/ecs.py
```diff
@@ -0,0 +1,153 @@
...
+            None,
+            lambda: self.ecs_client.run_task(**run_task_params),
+        )
+
+        if not response.get("tasks"):
+            failures = response.get("failures", [])
+            failure_details = "; ".join(
+                f"{f.get('reason', 'Unknown')}: {f.get('detail', 'No details')}" for f in failures
+            )
+            raise ClientError(
```
gusostow: Raising a botocore exception for an apsis error condition doesn't seem right.

python/apsis/program/procstar/aws/ecs_agent.py
```diff
@@ -0,0 +1,316 @@
...
+        self.timeout = timeout
+        self.mem_gb = mem_gb
+        self.vcpu = vcpu
+        self.disk_gb = disk_gb
+        self.role = role
+        # ECS programs don't have a predefined group_id - it's generated per-run
+        self.group_id = None
+
+    def __str__(self):
+        return self.argv[2] if len(self.argv) >= 3 else " ".join(self.argv)
```
gusostow: Just the third arg, not slicing up to the third?

gusostow (commented):
This is looking really solid. No real concerns here. 

I'm less worried about impact on traditional procstar program than I was originally.  No objections to rolling out refactor and ECS together, but up to you.

docs/programs.rst
LudovicoRighi: I'd have a slight preference in leaving as is, just to give the user a bit more flexibility.

python/apsis/program/procstar/aws/ecs.py
LudovicoRighi: Good idea.

docs/config.rst
LudovicoRighi: Removed, it's actually not required; we can just use the cluster's default.

python/apsis/program/procstar/aws/ecs.py
LudovicoRighi: No strong opinion either, but I've done that mostly because:
- `boto3` is always in sync with AWS API changes, whereas `aiobotocore` can occasionally lag behind
- avoid adding another third-party dependency to maintain

As far as I understand `aiobotocore` would be better for high-frequency and concurrent AWS calls; for our use case, the thread pool overhead is probably negligible.

gusostow: Makes sense.

docs/programs.rst
gusostow: ok your choice

gusostow (approved):
Ship it 📈

PR #520, opened 2026-02-11
Support parameterized `job_id` in dependencies

### LINKED ISSUE #324: parametrized dependencies
> Could we make conditions dependent on a parameter?

LudovicoRighi:
- Fixes: https://github.com/apsis-scheduler/apsis/issues/324

### Summary

- Support Jinja2 templates in dependency `job_id`, so a single job can depend on different jobs based on run parameters
- Add `enabled` field to all condition types, allowing conditions to be conditionally active per run
- Restructure the job detail page to show resolved dependencies inline with schedules

### Parameterized `job_id`

Both expression and block syntax are supported:

```yaml
condition:
  - type: dependency
    job_id: "region/{{ region }}"
  - type: dependency
    job_id: '{% if region == "us" %}dep/us{% else %}dep/eu{% endif %}'
```

Templates are expanded at bind time using the run's args. Non-string `job_id` values (e.g. unquoted `{{ }}` in YAML) are rejected with a helpful error.

### `enabled` field on conditions

```yaml
condition:
  - type: max_running
    count: 1
    enabled: "{{ strat == 'us' }}"            # only enable this for a specific strat
  - type: dependency
    job_id: download/my/data
    enabled: '{{ Date(date).weekday != 5 }}'      # disable/skip the dependency on a specific day of the week
```

- `true` or omitted — always active (backward compatible)
- `false` — unconditionally disabled
- `"{{ expr }}"` — evaluated at bind time; condition is skipped when false

Implemented in the `Condition` base class so all condition types (`dependency`, `skip_duplicate`, `max_running`) inherit it. When `bind()` returns `None`, callers in `runs.py` and `check.py` filter it out.

### Job page UI

The job detail page now shows resolved dependencies inline:
- Variable dependencies (those that differ across schedule arg sets) are shown per-group with resolved `job_id` and args
- Common dependencies are shown in a shared "Common Dependencies" section
- Conditions with `enabled=False` are omitted from the schedule display but visible in the collapsible "Definitions" section


python/apsis/cond/base.py
```diff
@@ -22,17 +22,59 @@ class Condition(TypedJso):
 
     TYPE_NAMES = TypedJso.TypeNames()
 
+    _enabled = None
+
+    @property
+    def enabled(self):
+        return self._enabled
+
+    @enabled.setter
```
gusostow: I'm surprised `Condition().enabled = ...` is used

python/apsis/lib/api.py
```diff
@@ -103,19 +104,115 @@ def _to_jsos(objs):
...
+                if entry is not None:
+                    resolved_groups[i]["conditions"].append(entry)
+
     return {
         "job_id": job.job_id,
         "params": list(sorted(job.params)),
         "schedule": [sched_to_jso(s) for s in job.schedules],
         "program": _to_jso(job.program),
         "condition": [_to_jso(c) for c in job.conds],
+        "common_conditions": common_conds_jso,
```
gusostow, line 238: Just noting that "condition" and "schedule" are singular. Not necessarily objecting because "common_condition" sounds weird.

python/apsis/lib/api.py
```diff
@@ -103,19 +104,115 @@ def _to_jsos(objs):
     return [] if objs is None else [_to_jso(o) for o in objs]
 
 
-def job_to_jso(job):
+def job_to_jso(job, jobs=None):
```
gusostow: Can you add a type hint and a docstring explaining why so much logic is required to prepare a job for serialization.

gusostow: Just flagging to make sure this isn't called in a tight loop anywhere, where performance hit could add up. Without looking that carefully it looks like this is becoming O(N^2) to do on a list of jobs

gusostow (commented):
Overall looking good. Will take UI conversation over DM.

python/apsis/cond/base.py
LudovicoRighi: I've refactored this: removed the `setter` and made `enabled` a private attribute

python/apsis/lib/api.py
LudovicoRighi: Yeah, noticed. I don't have a strong opinion about this, I'd leave as is unless you prefer the singular.

python/apsis/lib/api.py
LudovicoRighi: I've checked and added a complexity analysis on the docstring; this should be fine in terms of performance

gusostow (approved):

PR #525, opened 2026-03-03
Minor bugfix for `BoundSkipDuplicate`

LudovicoRighi:
Fix a bug in `BoundSkipDuplicate.__repr__` that referenced a nonexistent `self.__job_id` attribute, causing an `AttributeError` whenever `repr()` was called on an instance.


python/apsis/cond/skip_duplicate.py
```diff
@@ -95,7 +95,7 @@ def __repr__(self):
             self,
             check_states=self.__check_states,
             target_state=self.__target_state,
-            job_id=self.__job_id,
+            job_id=self.__run_id,
```
gusostow, line 98: I think you can get the job_id from self.__inst

PR #529, opened 2026-04-03
Allow specifying a yaml loader in load_jobs_dir

sjtechdev:
Adds a `yaml_loader` parameter to `load_jobs_dir` so callers can pass a faster PyYAML loaders, if they wish, instead of the default ruamel loader.


sjtechdev:
@LudovicoRighi @gusostow - this PR enables possibility for faster load_jobs_dir() calls; can chat more over DM.

gusostow (approved):

PR #530, opened 2026-04-23
Add timing and cycle checks for job dependencies

LudovicoRighi:
Extends `check_job_dependencies_scheduled` to catch two new classes of problems in job configurations:

**Timing checks:** When `max_wait_time` is set, the checker builds a dependency graph across all scheduled jobs and computes expected start times, accounting for transitive dependencies. If a job can't start until well after its scheduled time because a dependency (or a dependency's dependency) is scheduled too late, an error is reported with a tree-style explanation of the full blocking chain.

**Cycle detection:** The dependency graph is checked for cycles. If a cycle exists (e.g. A depends on B, B depends on C, C depends on A), errors are reported for all jobs involved, with the cycle path shown.

Both checks integrate into the existing `check-jobs` CLI flow and produce structured `(job, level, message)` tuples. 

(Also sneaked in `soft_wrap=True` to preserve tree indentation in the terminal)

### Example output

Timing violation:
```
A: A(date=2026-04-27) not expected to start until 2026-04-27T22:59:59Z, 24h after scheduled time 2026-04-26T23:00:01Z (max_wait=23h). This is because:
- dependency: B(date=2026-04-27) not expected to start until 2026-04-27T22:59:59Z. This is because:
  - dependency: C(date=2026-04-27) not expected to start until 2026-04-27T22:59:59Z. This is because:
    - It's scheduled to start at 2026-04-27T22:59:59Z.
```

Cycle detection:
```
A: dependency cycle detected: A → B → C → A
```

### DIFF
--- a/python/apsis/check.py
+++ b/python/apsis/check.py
@@ -1,3 +1,5 @@
+from collections import defaultdict, deque
+
 import ora
 
 from apsis.cond.dependency import Dependency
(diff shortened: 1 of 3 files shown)


gusostow (approved):
Nice

PR #531, opened 2026-06-12
Parse job yaml in a background threadpool

gusostow:
It turns out the event loop parses the thousands of yaml files back to back, which is taking 15s in our environment. If we have the bad luck of this running after a GC gen 2 collection then we start to have problems...

This does make loading jobs about 20% slower but that's fine as long as the event loop is happy.

The `asyncio.sleep(0)` is just to save from a subsequent few second hang on top of that.

### DIFF
--- a/python/apsis/jobs.py
+++ b/python/apsis/jobs.py
@@ -244,8 +244,6 @@ async def load_jobs_dir(path, yaml_loader=None):
     jobs = {}
     errors = []
 
-    loop = asyncio.get_running_loop()
-
     async def load_job(path, job_id):
         log.debug(f"loading: {path}")
         try:
@@ -259,7 +257,7 @@ async def load_jobs_dir(path, yaml_loader=None):
                     job_jso = YAML().load(content)
                 return Job.from_jso(job_jso, job_id)
 
-            job = await loop.run_in_executor(None, _parse)
+            job = await asyncio.to_thread(_parse)
             return job_id, job, None
         except DuplicateKeyError as exc:
             err_msg = exc.problem if exc.problem else str(exc)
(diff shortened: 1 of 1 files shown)


python/apsis/jobs.py
```diff
@@ -244,16 +244,22 @@ async def load_jobs_dir(path, yaml_loader=None):
...
-            job = Job.from_jso(job_jso, job_id)
+
+            def _parse():
+                if yaml_loader is not None:
+                    job_jso = yaml.load(content, Loader=yaml_loader)
+                else:
+                    job_jso = YAML().load(content)
+                return Job.from_jso(job_jso, job_id)
+
+            job = await loop.run_in_executor(None, _parse)
```
LudovicoRighi: I think `asyncio.to_thread` would be the slightly cleaner and more modern API here since we're using the loop's default `ThreadPoolExecutor`.

LudovicoRighi (approved):
Thanks!

Worth noting we explored this approach in:
- https://github.com/apsis-scheduler/apsis/pull/443#discussion_r1955162387 (085bc9b5 → reverted in 987cafd0) 

and concluded `to_thread` wasn't needed. In hindsight that conclusion was about parallelism/throughput; for event-loop responsiveness the off-loop yielding is exactly what we want, so re-landing seems the right call to me.


(Also, non-blocking, just noting for the future: the default `ThreadPoolExecutor` already caps concurrent parses (~32), so no thread-explosion risk. If we ever see a memory spike loading very large dirs, the lever would be a semaphore around read+parse (like #443 originally had) to bound in-flight file contents, but I wouldn't add it speculatively.)

gusostow:
> Worth noting we explored this approach in:

Yeah, my mistake. I thought I because the yaml parsing happening in separate tasks that it wouldn't starve the rest of apsis. I didn't realize that if they're all queued back to back they actually will.

PR #533, opened 2026-06-18
Remove bind at startup

gusostow:
This is a preliminary startup performance optimization for when the run store starts to query sqlite on-demand. It's not slow currently, but it will be slow after that change (20-30s).

I've queried our prod DB and I see that we have a 11 rows with null values here. Some of them are recent -- they're all jobs that failed to bind at runtime like adhoc runs, and some things that aren't caught by the jobs check.

I tested that nothing falls over without this bind at startup even if a few runs have nulls for actions/conds/program.

### DIFF
--- a/python/apsis/apsis.py
+++ b/python/apsis/apsis.py
@@ -101,21 +101,6 @@ class Apsis:
             min_timestamp = now() - lookback
         self.run_store = RunStore(db, min_timestamp=min_timestamp)
 
-        # Previously we didn't serialize conds or actions.  Bind them if
-        # missing on loaded runs.
-        # FIXME: Remove this after a while.
-        for run in self.run_store.query()[1]:
-            if run.actions is None or run.conds is None or run.program is None:
-                try:
-                    job = self.jobs.get_job(run.inst.job_id)
-                except LookupError:
-                    log.info(f"failed to bind {run}: job not found")
-                    continue
-                try:
-                    bind(run, job, self.jobs)
-                except Exception as exc:
-                    log.info(f"failed to bind {run}: {exc}")
-
         self.outputs = OutputStore(db.output_db)
 
         # Continue scheduling from the last time we handled scheduled jobs.


LudovicoRighi (approved):
This looks good to me to merge as-is. As far as I can tell the removed backfill was effectively inert:

- it ran in-memory only and never persisted, so it never durably fixed any row;
- the prod logs suggest it's just been re-failing to bind the same `error` run on each startup for over a month;
  ```
  > cat  apsis.log.{40..1} | grep "failed to bind"
  2026-05-30T15:26:18.244 apsis.apsis              I failed to bind r15762648 error .../test/run test script 1(): name 'sleep_time' is not defined
  2026-06-06T15:26:13.401 apsis.apsis              I failed to bind r15762648 error .../test/run test script 1(): name 'sleep_time' is not defined
  (...)
  2026-06-15T22:04:00.146 apsis.apsis              I failed to bind r15762648 error .../test/run test script 1(): name 'sleep_time' is not defined
  2026-06-16T21:14:39.187 apsis.apsis              I failed to bind r15762648 error .../test/run test script 1(): name 'sleep_time' is not defined
  ```
- and none of the existing null rows (`error`/`new`/`skipped`) look like they're ever engaged by `restore()` or the scheduler. 
  ```
  sqlite> SELECT state, COUNT(*) FROM runs WHERE program IS NULL OR conds IS NULL OR actions IS NULL GROUP BY state;
  ╭─────────┬──────────╮
  │  state  │ COUNT(*) │
  ╞═════════╪══════════╡
  │ error   │        7 │
  │ new     │        4 │
  │ skipped │        3 │
  ╰─────────┴──────────╯
  ```


One thing I wanted to raise for discussion (not a blocker): with the backfill gone, the invariant "an engaged run has non-null `program`/`conds`/`actions`" becomes load-bearing but unenforced. If a future bug ever did put a null on an active run, I think the failure mode could be pretty sharp:

- a `waiting`/`running` row would crash `restore()` → `SystemExit(1)`, so Apsis won't start;
- a future-dated `scheduled` row would pass restore silently and then crash the **live** scheduler loop at fire time.

The reason it escalates is that `_wait`/`__reconnect` run inside long-lived loops, so a `None` deref there takes down the service rather than just the run.

Would it be worth adding a small guard at those two entry points (top of `_wait`, after the asserts in `__reconnect`) that fails just the run? Something like:

```python
def _check_bound(self, run):
    if run.program is None or run.conds is None or run.actions is None:
        self.run_log.record(run, "not bound; cannot run")
        self._transition(run, State.error, force=True, times={"error": now()})
        return False
    return True
```

(I think `_start` wouldn't need a check since it's only reached via `_wait`.) 

Happy to be talked out of it, or to open it as a separate PR so this one stays a clean deletion — whatever you prefer. What do you think?

gusostow:
So this would be situations where run isn't bound but is able to be? Like some crash in between transitioning the run to `scheduled` and persisting the the bound run in in the DB? Right now I think that's just a single atomic update to the runs table.
 
But I suppose being defensive is good. You never know what would cause this. I'm going to merge for now though.

LudovicoRighi:
Yeah, correct. I was just suggesting to be as defensive as possible here, in case any future changes ever broke that atomicity.

PR #534, opened 2026-06-19
Add more filtering to RunDB

gusostow:
Performance preparations for loading RunStore data from sqlite instead of keeping in memory. This allows us to push filters down to sqlite instead of needing to deserialize all runs before filtering in memory.

This PR itself isn't risky because it's a new feature that doesn't change any runtime behavior. RunStore filtering will still be in-memory for now.

### DIFF
--- a/python/apsis/sqlite.py
+++ b/python/apsis/sqlite.py
@@ -495,8 +495,11 @@ class RunDB:
         with Timer() as timer:
             runs = list(self.__query_runs(self.__engine, expr))
 
+        def fmt_params(**kwargs):
+            return " ".join(f"{k}={v}" for k, v in kwargs.items() if v is not None)
+
         log.debug(
-            f"query job_id={job_id} since={since} state={state} "
+            f"query {fmt_params(run_ids=run_ids, job_id=job_id, since=since, state=state, args=args, with_args=with_args, min_timestamp=min_timestamp)} "
             f"→ {len(runs)} runs in {timer.elapsed:.3f}s"
         )
         return runs
--- a/python/apsis/sqlite.py
+++ b/python/apsis/sqlite.py
@@ -482,12 +482,16 @@ class RunDB:
         if args is not None:
             args = {str(k): str(v) for k, v in args.items()}
             for k, v in args.items():
-                where.append(sa.literal_column(f"json_extract(args, '$.{k}')") == v)
+                # escape key for JSON path to handle dots, quotes, etc
+                path = '$."' + k.replace('"', '\\"') + '"'
+                where.append(sa.func.json_extract(TBL_RUNS.c.args, path) == v)
             where.append(sa.literal_column("(SELECT count(*) FROM json_each(args))") == len(args))
         elif with_args is not None:
             with_args = {str(k): str(v) for k, v in with_args.items()}
             for k, v in with_args.items():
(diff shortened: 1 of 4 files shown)


python/apsis/sqlite.py
```diff
@@ -444,23 +445,59 @@ def get(self, run_id):
...
         return run
 
-    def query(self, *, job_id=None, since=None, min_timestamp=None):
+    def query(
+        self,
+        *,
+        run_ids=None,
+        job_id=None,
+        since=None,
+        state: Iterable[State] | Optional[State] = None,
```
LudovicoRighi: nit: I think the following would be more modern
```suggestion
        state: Iterable[State] | State | None = None,
```

python/apsis/sqlite.py
```diff
@@ -444,23 +445,59 @@ def get(self, run_id):
...
         if since is not None:
             where.append(TBL_RUNS.c.rowid >= int(since))
+        if state is not None:
+            states = py.tupleize(state)
+            where.append(TBL_RUNS.c.state.in_([s.name for s in states]))
+        if args is not None:
+            args = {str(k): str(v) for k, v in args.items()}
+            for k, v in args.items():
+                where.append(sa.literal_column(f"json_extract(args, '$.{k}')") == v)
+            where.append(sa.literal_column("(SELECT count(*) FROM json_each(args))") == len(args))
```
LudovicoRighi, line 488: Non-blocking and probably not reachable today, but worth a note: in the `args`/`with_args` predicates the arg key gets interpolated straight into the JSON path.

The value is bound, but the key isn't, and `.` is significant in JSON-path. So `{"a.b": "x"}` wouldn't match `query(args={"a.b": "x"})`: `$.a.b` reads as nested `a` → `b`, returns nothing, no error. (Quotes/spaces in keys are similar; they only work today because SQLite is lenient.)

The fix should be to quote the key and bind the path instead of building SQL text:

```python
path = '$."' + key.replace('"', '\\"') + '"'
sa.func.json_extract(TBL_RUNS.c.args, path) == value
```

And a test for the regression:

```python
r1 = _make_run(run_db, "job/a", {"a.b": "x"})
assert [r.run_id for r in run_db.query(args={"a.b": "x"})] == [r1.run_id]
```
(And maybe also something like:)
```python
_make_run(run_db, "job/a", {"name": "o'brien"})
_make_run(run_db, "job/a", {'wei"rd': "y"})
```

python/apsis/sqlite.py
```diff
@@ -444,23 +445,59 @@ def get(self, run_id):
...
             where.append(TBL_RUNS.c.timestamp >= dump_time(min_timestamp))
 
-        runs = list(self.__query_runs(self.__engine, sa.and_(*where)))
+        expr = sa.and_(*where)
+        with Timer() as timer:
+            runs = list(self.__query_runs(self.__engine, expr))
 
         log.debug(
-            f"query job_id={job_id} since={since} min_timestamp={min_timestamp} → {len(runs)} runs"
+            f"query job_id={job_id} since={since} state={state} "
```
LudovicoRighi: Any particular reason why you're dropping `min_timestamp`? I would rather add all the query params that you're adding for observability.

LudovicoRighi (approved):

python/apsis/sqlite.py
gusostow: Thanks nice catch

PR #536, opened 2026-06-21
Use run_summary cache table for websocket messages

gusostow:
Caches all summary json in sqlite to be streamed directly to the browser over websocket. This is a performance requirement for removing the in-memory run store.
 
I've updated the `migrate-db.py` script to create the table on existing DBs and backfill it.

TODO: is cleaning up the run summary table in the archive job. I'll do that in a follow up.

### DIFF
--- a/python/apsis/lib/api.py
+++ b/python/apsis/lib/api.py
@@ -244,11 +244,6 @@ def job_to_jso(job, jobs=None) -> dict:
 
 
 def run_to_summary_jso(run):
-    jso = run._summary_jso_cache
-    if jso is not None:
-        # Use the cached JSO.
-        return jso
-
     jso = {
         "job_id": run.inst.job_id,
         "args": run.inst.args,
@@ -264,8 +259,6 @@ def run_to_summary_jso(run):
         deps = [[c.job_id, c.args] for c in run.conds if isinstance(c, Dependency)]
         if len(deps) > 0:
             jso["dependencies"] = deps
-
-    run._summary_jso_cache = jso
     return jso
 
 
--- a/python/apsis/sqlite.py
+++ b/python/apsis/sqlite.py
@@ -518,7 +518,6 @@ class RunSummaryDB:
         "run_summary",
         METADATA,
         sa.Column("run_id", sa.String(), unique=True, nullable=False),
-        # approx run created timestamp for ordering return values, not guaranteed to match runs table
         sa.Column("timestamp", sa.Float, nullable=False),
         sa.Column("payload", sa.String(), nullable=False),
         # TODO: test if this index is necessary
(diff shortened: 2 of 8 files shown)


python/apsis/sqlite.py
```diff
@@ -512,6 +513,57 @@ def fmt_params(**kwargs):
...
+            """,
+            (run.run_id, dump_time(run.timestamp), payload),
+        )
+        self.__connection.connection.commit()
+
+    def query(self, min_timestamp: ora.Time) -> Iterator[str]:
+        with self.__engine.begin() as conn:
+            cursor = conn.execute(
+                sa.select(self.TABLE.c.payload)
+                .where(self.TABLE.c.timestamp >= dump_time(min_timestamp))
```
LudovicoRighi: What happens if `min_timestamp` is `None`? That could happen if `lookback` is not set in the apsis config:
https://github.com/apsis-scheduler/apsis/blob/a96b26c28da53869f1cc4aab84a3758c008efeb8/python/apsis/apsis.py#L96-L99

If we decide that `lookback` is mandatory to have in the config, then we should also modify the above ^; alternatively we have to handle the `None` case as well.

python/apsis/lib/api.py
```diff
@@ -244,11 +244,6 @@ def _resolve_cond(cond, sched_args):
 
 
 def run_to_summary_jso(run):
-    jso = run._summary_jso_cache
```
LudovicoRighi, line 247: Should we also remove the other 3 occurrences of `_summary_jso_cache` from `runs.py`?
https://github.com/apsis-scheduler/apsis/blob/a96b26c28da53869f1cc4aab84a3758c008efeb8/python/apsis/runs.py#L235

 I think that's dead code at this point

python/apsis/sqlite.py
```diff
@@ -512,6 +513,57 @@ def fmt_params(**kwargs):
...
+        payload = ujson.dumps(make_run_summary(run))
+        self.__connection.connection.execute(
+            """
+            INSERT INTO run_summary (
+               run_id,
+               timestamp,
+               payload
+            )
+            VALUES (?, ?, ?)
+            ON CONFLICT (run_id)
```
LudovicoRighi, line 545: Question about the `ON CONFLICT` clause here — I might be misreading, so worth a sanity check.

On update, this sets `payload` but leaves `timestamp` untouched, so `run_summary.timestamp` seems to stay pinned at the value from the first insert. Meanwhile `run.timestamp` gets refreshed to `now()` on every transition (`_transition` in `apsis.py`), and `RunDB.upsert` does write that fresh value back to `runs.timestamp` each time.

If that's right, the two tables' `timestamp` columns would drift apart over a run's lifetime. Since `summaries()` filters `run_summary` by `timestamp` while the in-memory store filters `runs` by its own `timestamp`, I *think* a run whose active span crosses the `min_timestamp` cutoff could end up loaded in memory but missing from the summary stream.

gusostow: Great catch thanks. For some reason I incorrectly assumed the timestamp in the runs table wasn't updated.

gusostow:
I addressed your comments. Merging without final re-review to get the maximum amount of time to run this change over the weekend. Glad to make any follow up fixes.

PR #538, opened 2026-07-01
RunStore: remove in-memory storage for most runs

gusostow:
This change is to avoid holding working set of lookback window runs in memory, which is contributing to horrendous GC gen 2  pauses.

The code changes in this PR are independent from run summary caching (https://github.com/apsis-scheduler/apsis/pull/536)  but performance is not acceptable until that is merged.

Terminology reminder: "expected" runs are in the scheduled state AND were created in the standard way from a schedule rather than an adhoc run e.g. a retry from the GUI. They historically haven't been persisted to the DB and they still aren't in this PR.


python/apsis/runs.py
LudovicoRighi: `with_args` is reassigned here from a dict (set at line 507) to a list of tuples. But the in-memory `expected` generator built at lines 508–509 is **lazy** — it isn't evaluated until `list(runs)` at line 538, by which point `with_args` is this list, and the generator calls `with_args.items()` on it:
```
 AttributeError: 'list' object has no attribute 'items'
```
 
 This fires on **any** `query(with_args=...)` call while ≥1 expected (scheduled) run is in memory — i.e. normal operation. It's reachable directly from the API: `GET /api/v1/runs?<arg>=<val>` routes extra query args into `with_args` (`service/api.py:489`), so the arg-filtered runs listing (CLI `apsis runs` with an arg filter, or the UI) 500s. Reproduced against the branch.
 

The post-chain re-filter block (lines 526–536) is also redundant: `since` aside, `args`/`with_args` are already applied to the expected generator (503–509) and pushed into the DB query (519–521). Recommend deleting the post-chain `args`/`with_args` re-filters entirely — that removes both the crash and the double work. (`since` does need handling — see Blocker/again below.)

python/apsis/runs.py
```diff
@@ -405,8 +407,11 @@ def update(self, run, timestamp):
...
-        # Make sure we know about this run.
-        assert self.__runs[run.run_id] is run
+        if run.state != State.new:
+            try:
+                _ = self.get(run.run_id)
+            except LookupError:
+                raise ValueError(f"unable to update unknown run: {run.run_id}")
 
         # Persist the changes, but not for expected runs.
         if not run.expected:
```
LudovicoRighi, line 412: `add()` puts a scheduled run in `__expected_runs` (line 399). When it later leaves `scheduled`, `Apsis._transition` sets `run.expected = False` and calls `update()`, which persists it to the DB here — but nothing ever removes it from `__expected_runs`. The only deletion path is `remove()`, called solely by `_unschedule_runs` for still-scheduled runs.
 
Result, reproduced on the branch through a normal lifecycle (scheduled→waiting→…→success):
- `query()` chains `__expected_runs` **and** the DB (line 519), so every transitioned run is returned **twice** (`run_ids=['r1','r1']`) — feeds `/runs`, the summary websocket, dependency checks.
- `__expected_runs` never shrinks, so the working set the PR set out to reclaim is exactly what leaks — this negates the PR's own GC-pause goal. `_retire_loop`/`retire()` used to be the eviction mechanism and are now gone/no-op. 

Fix: this is the natural chokepoint — when `not run.expected`, `self.__expected_runs.pop(run.run_id, None)` (it's now persisted and served from the DB).

python/apsis/sqlite.py
```diff
@@ -442,7 +486,10 @@ def upsert(self, run):
 
     def get(self, run_id):
         with self.__engine.begin() as conn:
-            (run,) = self.__query_runs(conn, TBL_RUNS.c.run_id == run_id)
+            run_or_none = list(self.__query_runs(conn, TBL_RUNS.c.run_id == run_id))
+            if len(run_or_none) == 0:
+                raise LookupError(f"no run: {run_id}")
```
LudovicoRighi, line 531: This PR changed `RunDB.get()` to raise `LookupError` (`sqlite.py:491`). `KeyError` is a subclass of `LookupError`, **not** the reverse — so this `except KeyError` 
https://github.com/apsis-scheduler/apsis/blob/fc51b6982734ac147cc908740f93802230aeb5cd/python/apsis/service/api.py#L201

will not catch it. For a missing/bogus run id, `run_store.get()` (which delegates to `RunDB.get()` for non-expected runs) now escapes the handler → **500 instead of 404**. Reproduced.

Same issue at line 247 (`/runs/<run_id>/updates` init). And these `get()` calls have **no guard at all**, so a missing run id now 500s: lines 356, 363, 371, 379, 391, 406, 425, 441.
 
Note the file is already inconsistent — 144, 159, 237, 312, 322, 338 were updated to `except LookupError`. The remaining `except KeyError` sites around `run_store.get()` need the same treatment, and the unguarded call sites need a guard.

python/apsis/runs.py
```diff
@@ -488,24 +491,37 @@ def query(
...
-        else:
-            # Slow path: scan.
-            runs = self.__runs.values()
+        if job_id is not None:
+            expected = (r for r in expected if r.inst.job_id == job_id)
 
-        if state is not None:
-            state = set(to_state(s) for s in iterize(state))
-            runs = (r for r in runs if r.state in state)
+        runs = itertools.chain(
```
LudovicoRighi: The docstring (line 483) states callers rely on ascending order, and `websocket_summary` does `runs[::-1]` for latest-first (`service/api.py:561-564`). This chains all in-memory `expected` runs first, then DB runs. Expected runs are the newest (just-scheduled) run_ids, so the concatenation is no longer globally ascending — reproduced `['r1','r3','r2','r4']` vs. `main`'s `['r1','r2','r3','r4']`. After the reversal, the summary ordering is wrong. Needs a real merge (sort combined by rowid/timestamp) or revisit the callers' assumption.

python/apsis/runs.py
```diff
@@ -488,24 +491,37 @@ def query(
...
+            expected,
+            self.__query_run_db(
+                run_ids=run_ids, job_id=job_id, state=state, args=args, with_args=with_args
+            ),
+        )
 
         if since is not None:
```
LudovicoRighi: `since` is not passed to `__query_run_db` (line 519-521) — it's filtered in Python after fetching, so a `since` query still deserializes the whole lookback window. `RunDB.query` already supports `since`; push it down.

python/apsis/runs.py
```diff
@@ -436,33 +441,31 @@ def retire(self, run_id):
           True if `run_id` is not in the store, either because it was
           successfully retired, or because it wasn't there to begin with.
         """
+        log.warning("no-op: RunStore.retire() deprecated")
```
LudovicoRighi: `retire()` is now a no-op that still returns the `state.finished` boolean `archive.py:132` relies on, but the name is misleading and it log-warns on every archived run (log spam). Consider renaming to `is_archivable(run_id)` and dropping the per-call warning.

python/apsis/runs.py
```diff
@@ -371,30 +372,31 @@ def __init__(self, db, *, min_timestamp):
...
-        # Populate cache from database.
-        self.__runs = {r.run_id: r for r in self.__run_db.query(min_timestamp=min_timestamp)}
-        # Keep a lookup of runs by job ID.
-        self.__runs_by_job = {}
-        for run in self.__runs.values():
-            self.__runs_by_job.setdefault(run.inst.job_id, set()).add(run)
+        self.__min_timestamp = min_timestamp
+
+        # in-memory storage of scheduled runs that came from a schedule ie not reruns or adhoc runs
+        # all runs are not expected unless they cam from the scheduler
```
LudovicoRighi:
```suggestion
        # all runs are not expected unless they came from the scheduler
```
nit: typo

python/apsis/runs.py
LudovicoRighi: Stale docstring: "Stores runs in memory … use `retire()` to retire older runs from memory" (359-362) no longer describes the design.

LudovicoRighi (commented):
I've also added some regression tests at:
- https://github.com/apsis-scheduler/apsis/pull/541

python/apsis/runs.py
gusostow: It doesn't actually need to be ascending anymore because the cached summaries are returned descending from the DB. Removed the outdated comment.

python/apsis/runs.py
gusostow: Woops I think that was from badly resolving a merge conflict. Thanks

python/apsis/service/api.py
```diff
@@ -481,6 +505,10 @@ async def runs(request):
     # were added to avoid collision with fixed args.
     args = {n[1:] if n.startswith("_") else n: a[-1] for n, a in args.items()}
 
+    # FIXME: add pagination. Without filters this query can return 1M runs and block the event loop for up to a 60s
+    if run_id is None and job_id is None and state is None and since is None and not args:
```
LudovicoRighi: nit: I think the following would be slightly more readable
```suggestion
    if all(x is None for x in (run_id, job_id, state, since)) and not args:
```

python/apsis/service/api.py
```diff
@@ -481,6 +505,10 @@ async def runs(request):
     # were added to avoid collision with fixed args.
     args = {n[1:] if n.startswith("_") else n: a[-1] for n, a in args.items()}
 
+    # FIXME: add pagination. Without filters this query can return 1M runs and block the event loop for up to a 60s
```
LudovicoRighi, line 509: Having at least one filter isn't really a guarantee here, right? A single *loose* filter passes the check but still matches the whole store — `since=1970-01-01`, `state=success`, or a common `arg` value all match ~every run. Only `run_id` genuinely bounds the result, so the event loop can still block on the case this guards against.

The cost is proportional to **rows returned** (per-row deserialize + serialize), not rows scanned — the SQL predicate itself is cheap even on a full scan. So a `LIMIT` in `RunDB.query` + a "truncated" indicator would bound it regardless of how loose the filter is, and is a natural step toward the pagination noted as the proper fix.

python/apsis/runs.py
```diff
@@ -423,47 +423,47 @@ def remove(self, run_id, *, expected=True):
...
         self.publisher.publish(self.Message(run.run_id, run.inst.job_id, run.inst.args, None))
         return run
 
-    def retire(self, run_id):
+    def archivable(self, run_id) -> bool:
         """
         :return:
-          True if `run_id` is not in the store, either because it was
-          successfully retired, or because it wasn't there to begin with.
+          True if `run_id` is not in the working set, either because it fished, or because it wasn't there to begin
```
LudovicoRighi:
```suggestion
          True if `run_id` is not in the working set, either because it finished, or because it wasn't there to begin
```
typo

python/apsis/cond/skip_duplicate.py
```diff
@@ -151,24 +151,30 @@ def from_jso(cls, jso):
...
+        # need run_id for the message; query to find the duplicate
         _, runs = run_store.query(
             job_id=self.__inst.job_id,
             args=self.__inst.args,
             state=self.__check_states,
         )
-        # Exclude this run itself.
         runs = [r for r in runs if r.run_id != self.__run_id]
 
         if len(runs) > 0:
```
LudovicoRighi: **`SkipDuplicate` can silently fail to skip a duplicate that's beyond the lookback window**

 

`RunStore.count_runs()` deliberately ignores `min_timestamp`, but `RunStore.query()` still applies it. `SkipDuplicate.check()` uses both — `count_runs()` to gate the fast path, then `query()` to locate the duplicate for the message. When a duplicate is in a check state (`waiting`/`starting`/`running`) but older than `now - lookback`, `count_runs()` sees it while `query()` can't, so `check()` returns `True` and the duplicate is **not** skipped.

 

Regression test (fails on this branch, passes on `main`): 
- https://github.com/apsis-scheduler/apsis/pull/543/changes

 

Severity is config-dependent — it's **latent, not reachable under our prod config** (`lookback=21d` ≫ `program.timeout=24.5h` / `waiting.max_time=23h`, so no run stays in a check state long enough). It becomes reachable if `lookback` is ever short relative to run lifetime. Worth fixing so correctness doesn't silently rely on that invariant: have the `query()` lookup in `check()` ignore `min_timestamp`, consistent with `count_runs()`.

python/apsis/service/api.py
gusostow: I'm just going to require either run_id or job_id for now, which is safe. This endpoint is barely used and I want to think about pagination more holistically when we tackle it.

python/apsis/runs.py
```diff
@@ -354,9 +355,8 @@ class RunStore:
     """
     Stores runs in memory.
```
LudovicoRighi: This is now misleading as most runs live in the DB; should we amend/remove it?

python/apsis/runs.py
```diff
@@ -423,47 +423,47 @@ def remove(self, run_id, *, expected=True):
...
+        else:
+            return True
 
     def get(self, run_id):
-        run = self.__runs[run_id]
+        if run_id in self.__expected_runs:
+            try:
+                run = self.__expected_runs[run_id]
+            except KeyError:
+                raise LookupError(f"no run: {run_id}")
```
LudovicoRighi: Is this reachable?

python/apsis/cond/skip_duplicate.py
gusostow: I added an optional override to ignore the min_timestamp

python/apsis/runs.py
```diff
@@ -405,11 +403,12 @@ def update(self, run, timestamp):
...
         """
-        # Make sure we know about this run.
-        assert self.__runs[run.run_id] is run
-
         # Persist the changes, but not for expected runs.
         if not run.expected:
+            try:
+                self.__expected_runs.pop(run.run_id)
+            except KeyError:
+                pass
```
LudovicoRighi: `update()` pops the run from `__expected_runs` *before* the DB upserts. If either upsert raises (DB locked, disk full, ...), the run has already left memory but isn't in the DB → it's dropped from both → silent data loss. This is a regression vs. main, where a run was never removed from the in-memory set, so a failed upsert left it recoverable.

 

Suggest upserting **before** popping:
```python
        if not run.expected:
            self.__run_db.upsert(run)
            self.__summary_db.upsert(run)
            try:
                self.__expected_runs.pop(run.run_id)
            except KeyError:
                pass
```

 

This trades unrecoverable loss for, at worst, a transient duplicate (if `run_summary` fails after `run_db` succeeds the run isn't popped, so it shows up in both until the next successful update, or until restart on a terminal eviction). A duplicate is observable and non-destructive, so it's the better failure mode; fully removing the window would need both writes + the pop in one transaction.

 

Regression test (fails on the current ordering, passes on main and with the reorder): 
- https://github.com/apsis-scheduler/apsis/pull/545#issuecomment-4980691786

-------

Separately, small nit:
```python
            try:
                self.__expected_runs.pop(run.run_id)
            except KeyError:
                pass
```
can be written more idiomatically as
```python
                self.__expected_runs.pop(run.run_id, None)
```

LudovicoRighi: Also, maybe worth a comment on why we're ignoring `KeyError`? My understanding is that this is because the unexpected runs (i.e. reruns, adhoc runs) are never inserted into `__expected_runs`, correct?

gusostow: That pop happens on every update but the run is only in that dict when it was expected and transitioning out of the scheduled state.

python/apsis/sqlite.py
```diff
@@ -443,15 +484,17 @@ def upsert(self, run):
 
     def get(self, run_id):
         with self.__engine.begin() as conn:
-            (run,) = self.__query_runs(conn, TBL_RUNS.c.run_id == run_id)
+            run_or_none = list(self.__query_runs(conn, TBL_RUNS.c.run_id == run_id))
```
LudovicoRighi: **`RunDB.get()` and `run_ids=` queries do a full table SCAN.**

The `runs` table has no index on `run_id` — only `index_runs_job_id` and the `rowid`
primary key. So `WHERE run_id = ?` (and `WHERE run_id IN (...)`) can't use an index.
Verified against the prod DB (938,541 runs):

On `main`, `RunStore.get()` was an O(1) in-memory dict lookup. This PR routes it to
`RunDB.get()`, so every `run_id` access now scans ~940k rows on the event loop (none of
these are `to_thread`'d). 

**Fix (no migration needed):** `run_id` is just `"r" + rowid`, and `rowid` *is* the
indexed primary key — `_parse_run_id()` already converts one to the other. Filter on
`rowid` instead of `run_id` in `get()` and in the `run_ids` branch of `__build_where`
(`rowid IN [_parse_run_id(i) for i in run_ids]`). That turns the SCAN into a PK SEARCH
(~40 ms → ~0.03 ms) with zero schema change.

https://github.com/apsis-scheduler/apsis/blob/d270d2576d76ddd1caa5cc1bf3b5556ab4a8c7de/python/apsis/sqlite.py#L60-L63

Alternatively add `CREATE INDEX ... ON runs (run_id)`, but that needs a migration on the
938k-row prod table and an extra index to maintain, when the rowid PK is already there for
free.


```
sqlite> SELECT name, sql FROM sqlite_master WHERE type='index' AND tbl_name='runs';
╭───────────────────┬─────────────────────────────────────────────────╮
│       name        │                       sql                       │
╞═══════════════════╪═════════════════════════════════════════════════╡
│ index_runs_job_id │ CREATE INDEX index_runs_job_id ON runs (job_id) │
╰───────────────────┴─────────────────────────────────────────────────╯
sqlite> EXPLAIN QUERY PLAN SELECT * FROM runs WHERE run_id = 'r5';
QUERY PLAN
`--SCAN runs
sqlite> EXPLAIN QUERY PLAN SELECT * FROM runs WHERE rowid = '5';
QUERY PLAN
`--SEARCH runs USING INTEGER PRIMARY KEY (rowid=?)
```

```
sqlite> SELECT run_id, rowid from runs limit 5;
╭──────────┬─────────╮
│  run_id  │  rowid  │
╞══════════╪═════════╡
│ r5438393 │ 5438393 │
│ r5438394 │ 5438394 │
│ r5438395 │ 5438395 │
│ r5438396 │ 5438396 │
│ r7366775 │ 7366775 │
╰──────────┴─────────╯
```

gusostow: Nice catch thanks. Filtering on rowid now.

gusostow:
Shipping it. 

$63 dollars later the Mythos super weapon is happy https://gist.github.com/gusostow/d3ec92f1f8bb503e4c64edd83d2a15aa. 

Note that I addressed one issue after that by erroring out old zombie runs that are outside the lookback window in the DB migration script I'm going to run https://github.com/apsis-scheduler/apsis/pull/538/changes/47b905763d07847111fba77dffb7b078156367d0.

PR #540, opened 2026-07-10
Add regression test for the `GET /api/v1/runs` endpoint

LudovicoRighi:
#539 moved `/runs` serialization off the event loop into a background thread (`asyncio.to_thread`) to stop it blocking the loop on large, unfiltered queries. This fixes the blocking, but introduces a data race.
 

### Why #539 is broken

Before #539, `runs_to_jso` ran synchronously on the event loop. It contains no `await`, so it was atomic with respect to every other coroutine — nothing could mutate a run mid-serialization. Slow, but safe.

 
Under `to_thread`, serialization runs concurrently with the event loop, which keeps driving the run state machine (`_process_updates` → `Apsis._transition`) over the *same* `Run` objects being serialized. `run_to_summary_jso` iterates `run.times.items()` while `Run._transition` mutates that dict in place (`self.times[state.name] = ...`). The result:

```
RuntimeError: dictionary changed size during iteration
```

 
The likelihood scales with exactly the case #539 targets — many runs → long serialization window → more transitions land inside it — so it's not a corner case. (`run.meta` is aliased into the JSO and also mutated in place by `_transition`, so it's exposed the same way.)

 
### The test

`test/unit/test_runs_endpoint_race.py` drives the real `api.runs` handler in process (mock DB — no server, network, or real Apsis) with a background task transitioning runs, and asserts the handler never crashes. 

- **Fails on current `main`** with the `RuntimeError` above.
- **Passes if 80a547a is reverted.**

The race is probabilistic per call, so the test serves a large run set and repeats the call several times; at these sizes a single buggy call crashes with probability well above 1/2, making a false pass negligible while a correct (atomic) implementation passes every call. It reproduces at Python's default thread-switch interval — no scheduler tuning.

### Scope

This PR adds only the failing test to pin down the regression. The fix is deliberately left to follow-up discussion.

cc @gusostow

### DIFF
--- a/python/apsis/service/api.py
+++ b/python/apsis/service/api.py
@@ -489,16 +489,7 @@ async def runs(request):
         with_args=args,
     )
 
-    # offload JSON serialization to thread pool to avoid blocking event loop
-    # runs is a generator, so materialize it first in the async context
-    runs_list = list(runs)
-
-    def serialize_runs():
-        return runs_to_jso(request.app, when, runs_list, summary=summary)
-
-    result = await asyncio.to_thread(serialize_runs)
-
-    return response_json(result)
+    return response_json(runs_to_jso(request.app, when, runs, summary=summary))
 
 
 async def _send_chunked(msgs, ws, prefix):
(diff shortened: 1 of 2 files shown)


LudovicoRighi:
The test fails on latest `main` (which includes 80a547a)
```
FAILED test/unit/test_runs_endpoint_race.py::test_runs_endpoint_survives_concurrent_transitions - AssertionError: /runs crashed under concurrent transitions: RuntimeError('dictionary changed size during iteration')
assert RuntimeError('dictionary changed size during iteration') is None
```

LudovicoRighi:
And passes after reverting https://github.com/apsis-scheduler/apsis/commit/80a547a40fab741bf45ac2782be5f686d1de4489

PR #551, opened 2026-08-03
Bugfix: fix broken `schedule.max_age` check

LudovicoRighi:
## Problem

`schedule.max_age` is unusable, and has been since 80044203 (2022-11-01).

The check at `scheduler.py:120` reads `self.stop`, but the attribute is the name-mangled `self.__stop`; no `stop` property exists. Every scheduler loop iteration raises `AttributeError`, which the loop's `except Exception` converts to `SystemExit(1)`.

The effect is worse than the guard not working: **setting `schedule.max_age` at all makes Apsis unbootable.** Confirmed against a real `apsisctl serve` — the process logs `scheduler loop failed` / `AttributeError: 'Scheduler' object has no attribute 'stop'` and is dead within seconds of startup.

This survived three years because `max_age` defaults to `None` and no config in the repo sets it, so the buggy expression short-circuits before reaching the bad attribute. Nothing in production is currently affected — this is "the documented safety valve can't be turned on," not "prod is fragile."

## Change

`self.stop` → `self.__stop`.

`git log -S` shows `self.stop` entered in exactly one commit, the "Reorg schedule config" refactor; the line before it read `self.__stop`. That refactor also rewrote the comparison from `time - stop > max_age` to `max_age < time - stop`, which is the same predicate (verified across values including the exact boundary). So this restores the original behavior rather than guessing at intent.

### DIFF
--- a/python/apsis/scheduler.py
+++ b/python/apsis/scheduler.py
@@ -117,7 +117,7 @@ class Scheduler:
                 time = now()
                 log.debug(f"scheduler loop: {time}")
 
-                if self.__max_age is not None and self.__max_age < time - self.stop:
+                if self.__max_age is not None and self.__max_age < time - self.__stop:
                     raise RuntimeError(f"last scheduled more than {self.__max_age} s ago")
 
                 await self.schedule(time + self.__horizon)
--- a/test/unit/test_scheduler.py
+++ b/test/unit/test_scheduler.py
@@ -1,0 +1,126 @@
+import asyncio
+import ora
+import pytest
+
+from apsis.scheduler import Scheduler
+
+# -------------------------------------------------------------------------------
+
+
+class _PollSleep(BaseException):
+    """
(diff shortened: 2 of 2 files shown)


LudovicoRighi:
As expected, regression tests fail before the fix.
```
=========================== short test summary info ============================
FAILED test/unit/test_scheduler.py::test_max_age_ok - SystemExit: 1
FAILED test/unit/test_scheduler.py::test_max_age_within - SystemExit: 1
FAILED test/unit/test_scheduler.py::test_max_age_exceeded - AssertionError: unexpected cause: AttributeError("'Scheduler' object has no attribute 'stop'")
assert False
 +  where False = isinstance(AttributeError("'Scheduler' object has no attribute 'stop'"), RuntimeError)
FAILED test/unit/test_scheduler.py::test_scheduling_advances_past_max_age - SystemExit: 1
```

gusostow (approved):

PR #553, opened 2026-08-09
Parse job files with a fast libyaml loader

gusostow:
Stacked on top of #557.

Replace ruamel's `YAML().load` with `DupCheckSafeLoader`, a libyaml (`yaml.CSafeLoader`) subclass, and parse each file in a thread. Preserves ruamel's behavior: duplicate-key rejection, YAML 1.2 scalar rules, merge keys, empty-scalar-as-null. ~24x faster parse; malformed YAML is now a per-job error instead of aborting the reload.

Review only the commit on top of #557.

### DIFF
--- a/python/apsis/jobs.py
+++ b/python/apsis/jobs.py
@@ -4,10 +4,10 @@ import logging
 import os
 from pathlib import Path
 import random
+import re
 import string
 import yaml
 from ruamel.yaml import YAML
-from ruamel.yaml.constructor import DuplicateKeyError
 
 from .actions import Action
 from .actions.schedule import successor_from_jso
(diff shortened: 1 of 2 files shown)


python/apsis/jobs.py
```diff
@@ -270,14 +266,15 @@ def _parse():
...
-                errors.append(exc)
+    # Load one file at a time, yielding to the loop after each, so a reload
+    # doesn't block the event loop.
+    for path, job_id in list_yaml_files(jobs_path):
+        _, job, exc = await load_job(path, job_id)
+        if job is not None:
+            jobs[job_id] = job
+        if exc is not None:
+            errors.append(exc)
+        await asyncio.sleep(0)
```
gusostow: Is the sleep necessary here? I would only think we need to give back to the event loop deliberately if we were scheduling large number of tasks all at once with gather or something like that.

python/apsis/jobs.py
```diff
@@ -270,14 +266,15 @@ def _parse():
...
-        results = await asyncio.gather(*chunk)
-        for job_id, job, exc in results:
-            if job is not None:
-                jobs[job_id] = job
-            if exc is not None:
-                errors.append(exc)
+    # Load one file at a time, yielding to the loop after each, so a reload
+    # doesn't block the event loop.
+    for path, job_id in list_yaml_files(jobs_path):
+        _, job, exc = await load_job(path, job_id)
```
gusostow, line 347: I've lost the description you gave me over slack that had benchmark numbers. Is this slower overall because you're parsing the yaml sequentially instead of in a threadpool? 

Granted that performance doesn't matter too much here as long as we're not blocking the event loop, which we don't seem to be.

python/apsis/jobs.py
```diff
@@ -270,14 +266,15 @@ def _parse():
...
 
-    load_coros = [load_job(path, job_id) for path, job_id in list_yaml_files(jobs_path)]
-    for chunk in itr.chunks(load_coros, 100):
-        results = await asyncio.gather(*chunk)
-        for job_id, job, exc in results:
-            if job is not None:
-                jobs[job_id] = job
-            if exc is not None:
-                errors.append(exc)
+    # Load one file at a time, yielding to the loop after each, so a reload
```
gusostow: You could tighten up this comment. It's not the number of jobs loaded at a time that was blocking the event loop. It was that there was a synchronous walk of the jobs directory, which this now does lazily.

python/apsis/jobs.py
```diff
@@ -155,9 +154,87 @@ def dump_yaml(file, job):
     YAML().dump(job_to_jso(job), file)
 
 
+class DuplicateKeyError(Exception):
+    """A YAML mapping contains a duplicate key."""
+
+
+class _DupCheckSafeLoader(yaml.CSafeLoader):
```
gusostow, line 161: Remind me, what's the performance payoff of using this custom loader? Tweaking it to this extent makes me a bit anxious. 

Remember, absolute job load performance isn't the top priority here. It's more not blocking the event loop and being 100% correct.

python/apsis/jobs.py
```diff
@@ -155,9 +154,87 @@ def dump_yaml(file, job):
...
+
+
+DupCheckSafeLoader = _build_yaml_loader()
+
+
 def list_yaml_files(dir_path):
     dir_path = Path(dir_path)
-    for dir, _, names in os.walk(dir_path):
+    for dir, dirs, names in os.walk(dir_path):
+        # Don't go into hidden dirs (e.g. `.git`)
```
gusostow, line 236: Nice

gusostow (commented):
Congrats on the first apsis contribution @fanman1

python/apsis/jobs.py
gusostow: Never mind. I accidentally commented on an old commit

python/apsis/jobs.py
```diff
@@ -252,32 +329,27 @@ async def load_job(path, job_id):
...
 
-    load_coros = [load_job(path, job_id) for path, job_id in list_yaml_files(jobs_path)]
-    for chunk in itr.chunks(load_coros, 100):
-        results = await asyncio.gather(*chunk)
-        for job_id, job, exc in results:
-            if job is not None:
-                jobs[job_id] = job
-            if exc is not None:
-                errors.append(exc)
+    # Load one file at a time. `load_job` parses in a thread, which yields to the loop, so a reload doesn't block it.
```
gusostow: You could tighten up this comment. It's not the number of jobs loaded at a time that was blocking the event loop. It was that there was a synchronous walk of the jobs directory, which this now does lazily.

PR #554, opened 2026-08-12
Bugfix: don't drop run metadata when a program result has none

LudovicoRighi:
## Problem

A run can lose all of its program metadata, most visibly when it transitions to `error`.

Program metadata is applied as `run.meta.update({"program": msg.meta})`, which *replaces* the whole `program` value, and every program message defaults to `meta={}`. So a message that reports no metadata erases everything already known about the program. Three paths do exactly that: `reconnect failed` (`agent.py:549`), the generic `procstar: {exc}` handler when no result has arrived yet (`agent.py:678`), and `start failed` (`agent.py:515`, which for ECS erases the pre-start `aws_ecs` record).

Reproduced end to end — start a run, stop Apsis, restart the agent with a new connection ID, start Apsis. Apsis can't reconnect, errors the run, and `meta` becomes `{'job': {'labels': []}, 'program': {}}`, losing the agent host, connection ID, proc ID and pid: precisely what's needed to go find the process Apsis just lost track of.

## Fix

Distinguish "reported no metadata" from "reported empty metadata", and produce no update in the former case:

```python
def _program_meta(meta):
    return {"program": meta} if meta else {}
```

applied at the seven sites in `apsis/running.py` that turn a program message into a metadata update. For non-empty metadata the result is identical to before, so success, failure, intermediate results and ECS base metadata are unaffected; replace-vs-merge semantics are deliberately left alone. This also unifies the pre-existing `if update.meta is not None` guard, which the running loop had but the startup loop didn't.

## One behaviour change

For programs that never report metadata (`no-op`, internal `stats`), `meta.program` is now absent rather than `{}`. All in-repo consumers guard for this (`RunView.vue:79`, and no Python code indexes `meta["program"]` outside `running.py`), but an external consumer doing `run["meta"]["program"]` would need `.get()`.

## Tests

- `test/unit/test_running.py` drives the real `_process_updates` over fixed message sequences against a real `Run` and the real `Apsis._transition`/`_update_metadata`: a metadata-free `ProgramSuccess`/`ProgramFailure`/`ProgramError` keeps the metadata, an error during startup keeps `_pre_start` metadata, and a result *with* metadata still replaces it.
- `test/int/procstar/test_program.py::test_reconnect_failed_keeps_metadata` covers the reported failure for real.

With the `running.py` change reverted, 4 of the 6 unit tests and the integration test fail. Full suite: 491 passed (`test_stop.py::test_rerun_with_stop` is flaky on `main` too, on its `elapsed < 0.1` timing assertion).

## Not addressed

Sub-key loss under replace semantics: keys a later result omits still disappear — e.g. `proc_statm` is present while running and absent once terminated, since Procstar can't read `/proc/<pid>/statm` for a dead process. Merging would retain those but also keep stale snapshots on finished runs.

### DIFF
--- a/python/apsis/running.py
+++ b/python/apsis/running.py
@@ -46,6 +46,19 @@ async def _maybe_compress(outputs, *, compression="br", min_size=16384):
     return dict(zip(outputs.keys(), o))
 
 
+def _program_meta(meta):
+    """
+    Returns a run metadata update for program metadata `meta`.
+
+    A run's program metadata is replaced by, rather than merged with, the
+    metadata of each program update, so a program that reports no metadata must
+    produce no update at all.  Otherwise it would erase what is already known
+    about the program, such as the host and pid of a process that Apsis has
+    lost track of.
+    """
+    return {"program": meta} if meta else {}
+
+
 async def _process_updates(apsis, run):
     """
     Processes program `updates` for `run` until the program is finished.
@@ -65,7 +78,7 @@ async def _process_updates(apsis, run):
                     # Handle intermediate updates during startup
                     if update.outputs is not None:
                         apsis._update_output_data(run, update.outputs, False)
-                    apsis._update_metadata(run, {"program": update.meta})
+                    apsis._update_metadata(run, _program_meta(update.meta))
 
                 case ProgramRunning() as running:
                     apsis.run_log.record(run, "running")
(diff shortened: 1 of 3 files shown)


gusostow (approved):
Nice fix!

PR #555, opened 2026-09-02
expose run args to its process as APSIS_ARG_* env vars

fanman1:
## Problem
A run's args (like `date` or `database`) were only reachable by interpolating them into the command with `{{ }}`; there was no way for a process to read them as environment variables. That is inconvenient for jobs that need an arg deep in their own code rather than on the command line, for example threading the scheduled time through medallion download functions ([asd/data#3657](https://git.tudor.com/asd/data/issues/3657)).

## Fix
After binding, `runs.bind` passes the run's args to a new `Program.set_run_args` method. The procstar bound programs store them, and their shared proc spec emits each as `APSIS_ARG_<name>` next to `APSIS_RUN_ID`, so both agent and ECS jobs are covered. Values are stringified and persisted in `to_jso`/`from_jso` so they survive restart and reconnect.

## Tests
- `date` and `database` emitted as `APSIS_ARG_*`, including args not referenced in the command
- shell (`ProcstarShellProgram`) and argv (`ProcstarProgram`) programs
- full `runs.bind` path with a real Job/Instance/Run
- ECS emission through the shared proc spec
- args survive a JSO round trip and are still emitted on reconnect
- non-string values stringified (`5`, `1.5`)
- `set_run_args` overwrites previously set args
- an arg named `RUN_ID` does not clobber `APSIS_RUN_ID`
- empty string value
- no args emits only `APSIS_RUN_ID`
- realistic param names from real jobs (`date`, `time`, `daytime`, `instance`, `strat`, `macro_portfolio`) plus digit/underscore names
- base `set_run_args` is a no-op for non-procstar programs

### DIFF
--- a/python/apsis/program/base.py
+++ b/python/apsis/program/base.py
@@ -15,6 +15,14 @@ log = logging.getLogger(__name__)
 
 TIMEOUT_SIGNAL = Signals.SIGTERM.name
 
+# Prefix for environment variables carrying a run's bound args to its process.
+APSIS_ARG_ENV_PREFIX = "APSIS_ARG_"
+
+
+def normalize_args(args):
+    """Returns an args mapping as a `{str: str}` dict."""
+    return {str(k): str(v) for k, v in (args or {}).items()}
+
 # -------------------------------------------------------------------------------
 
 
@@ -254,6 +262,12 @@ class Program(TypedJso):
     def bind(self, args):
         """
         Returns a new program with parameters bound to `args`.
+        """
+
+    def set_run_args(self, args):
+        """
+        Records the run's bound args on the bound program.  Override to expose
+        them to the process; by default does nothing.
         """
 
     # FIXME: Find a better way to get run_id into logging without passing it in.
--- a/python/apsis/program/base.py
+++ b/python/apsis/program/base.py
@@ -15,7 +15,6 @@ log = logging.getLogger(__name__)
 
 TIMEOUT_SIGNAL = Signals.SIGTERM.name
 
-# Prefix for environment variables carrying a run's bound args to its process.
 APSIS_ARG_ENV_PREFIX = "APSIS_ARG_"
 
 
@@ -196,8 +195,6 @@ class RunningProgram:
 
     def __init__(self, run_id):
         self.run_id = run_id
-        # The run's bound args, applied via `set_run_args` before the process
-        # starts, for exposing to it.
         self.args = {}
 
     def set_run_args(self, args):
--- a/python/apsis/program/base.py
+++ b/python/apsis/program/base.py
@@ -23,6 +23,7 @@ def normalize_args(args):
     """Returns an args mapping as a `{str: str}` dict."""
     return {str(k): str(v) for k, v in (args or {}).items()}
 
+
 # -------------------------------------------------------------------------------
 
 
--- a/python/apsis/program/base.py
+++ b/python/apsis/program/base.py
@@ -196,6 +196,15 @@ class RunningProgram:
 
     def __init__(self, run_id):
         self.run_id = run_id
+        # The run's bound args, applied via `set_run_args` before the process
+        # starts, for exposing to it.
+        self.args = {}
+
+    def set_run_args(self, args):
+        """
+        Records the run's bound args, for exposing to the process.
+        """
+        self.args = normalize_args(args)
 
     @property
     def updates(self):
@@ -263,12 +272,6 @@ class Program(TypedJso):
     def bind(self, args):
         """
         Returns a new program with parameters bound to `args`.
-        """
-
-    def set_run_args(self, args):
-        """
-        Records the run's bound args on the bound program.  Override to expose
-        them to the process; by default does nothing.
         """
 
     # FIXME: Find a better way to get run_id into logging without passing it in.
--- a/python/apsis/program/base.py
+++ b/python/apsis/program/base.py
@@ -17,12 +17,6 @@ TIMEOUT_SIGNAL = Signals.SIGTERM.name
 
 APSIS_ARG_ENV_PREFIX = "APSIS_ARG_"
 
-
-def normalize_args(args):
-    """Returns an args mapping as a `{str: str}` dict."""
-    return {str(k): str(v) for k, v in (args or {}).items()}
-
-
 # -------------------------------------------------------------------------------
 
 
@@ -196,12 +190,6 @@ class RunningProgram:
     def __init__(self, run_id):
         self.run_id = run_id
         self.args = {}
-
-    def set_run_args(self, args):
-        """
-        Records the run's bound args, for exposing to the process.
-        """
-        self.args = normalize_args(args)
 
     @property
     def updates(self):
--- a/python/apsis/program/procstar/agent.py
+++ b/python/apsis/program/procstar/agent.py
@@ -21,11 +21,13 @@ from apsis.lib.py import or_none, nstr, get_cfg
 from apsis.procstar import get_agent_server
 from apsis.program import base
 from apsis.program.base import (
+    APSIS_ARG_ENV_PREFIX,
     ProgramSuccess,
     ProgramFailure,
     ProgramError,
     Timeout,
     get_global_runtime_timeout,
+    normalize_args,
 )
 from apsis.program.process import Stop, BoundStop
 from apsis.runs import join_args, template_expand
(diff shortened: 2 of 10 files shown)


python/apsis/program/procstar/agent.py
```diff
@@ -458,8 +468,12 @@ def _spec(self):
         return procstar.spec.Proc(
             self._spec_argv,
             env=procstar.spec.Proc.Env(
+                # APSIS_RUN_ID is set first, then the run's args; the ordering
+                # keeps args from shadowing it.  Arg names are trusted (not
+                # validated as env-var identifiers).
```
LudovicoRighi: We shouldn't rely on ordering, that's not robust enough. I think we're actually relying on the `APSIS_ARG_ENV_PREFIX`, aren't we?

python/apsis/program/procstar/agent.py
```diff
@@ -370,6 +377,7 @@ def to_jso(self):
             | if_not_none("sudo_user", self.sudo_user)
             | ifkey("resources", self.resources.to_jso(), {})
             | ifkey("stop", self.stop.to_jso(), {})
+            | ifkey("args", self.args, {})
```
LudovicoRighi: `args` in the bound program's `to_jso()` duplicates what `runs.args` already holds, and it's a key that pre-feature Apsis rejects — I rolled back against a DB written by this branch and `restore()` died with `SchemaError: unexpected keys: args` → `SystemExit(1)`.

Is the persistence needed? `run.inst.args` is on the Run at every point a program gets started, so I'm not sure what the JSO copy buys us.

python/apsis/program/procstar/aws/ecs_agent.py
```diff
@@ -174,6 +175,7 @@ def __init__(
         disk_gb: Optional[int] = None,
         role: Optional[str] = None,
         task_definition: Optional[str] = None,
+        args=None,
```
LudovicoRighi: Is this a dead constructor parameter?

python/apsis/program/procstar/aws/ecs_agent.py
LudovicoRighi, line 1: I believe we should be able to implement this feature in the base class (maybe `RunningProgram`?) as I believe this is a feature that we want to support for all kind of programs. Have you considered this / are there any blockers?
(I'd expect no changes to be required in this file.)

python/apsis/program/procstar/agent.py
```diff
@@ -349,17 +351,22 @@ def __init__(
         stop=BoundStop(),
         timeout=None,
         resources=BoundResources(),
+        args=None,
```
LudovicoRighi: Is this a dead constructor parameter?

python/apsis/program/procstar/aws/ecs_agent.py
fanman1: thanks for the catch, I added `args` to `RunningProgram` since the dictionary is filled in at `_start` with `set_run_args` and then piped into the env variables at `_spec`.

python/apsis/program/procstar/agent.py
```diff
@@ -43,13 +44,16 @@
...
 
 
-def _sudo_wrap(cfg, argv, sudo_user):
+def _sudo_wrap(cfg, argv, sudo_user, arg_env_names=()):
     if sudo_user is None:
         return argv
     else:
         sudo_argv = get_cfg(cfg, "sudo.argv", SUDO_ARGV_DEFAULT)
+        # sudo scrubs env; extend its --preserve-env allowlist with this run's args.
+        preserve = [f"--preserve-env={','.join(arg_env_names)}"] if arg_env_names else []
```
LudovicoRighi: I believe we shouldn't need to use `--preserve-env` here. Please notice this would currently break all our prod mod jobs.

Which env vars we're allowed to preserve when sudo-ing is determined by our sudo configs. We already use `env_keep` to preserve `APSIS_RUN_ID` and `PROCSTAR_AGENT_HOST`. I believe we should be able to preserve also `APSIS_ARG_*`. More details internally.

python/apsis/program/base.py
```diff
@@ -15,6 +15,14 @@
 
 TIMEOUT_SIGNAL = Signals.SIGTERM.name
 
+APSIS_ARG_ENV_PREFIX = "APSIS_ARG_"
+
+
+def normalize_args(args):
```
LudovicoRighi: `args` should already be strings at this point. I believe we may be able to remove this function (i.e. it's a no-op), but please verify carefully.

Quick testing:
- Modified the function to just return `args` as is:
  ```diff
  > git diff
  diff --git a/python/apsis/program/base.py b/python/apsis/program/base.py
  index aa2e8902..54cc278c 100644
  --- a/python/apsis/program/base.py
  +++ b/python/apsis/program/base.py
  @@ -20,7 +20,7 @@ APSIS_ARG_ENV_PREFIX = "APSIS_ARG_"
  
   def normalize_args(args):
       """Returns an args mapping as a `{str: str}` dict."""
  -    return {str(k): str(v) for k, v in (args or {}).items()}
  +    return args
  
  
   # -------------------------------------------------------------------------------
  ```
 
- Test still passes:
  ```
  > pytest test/unit/procstar/test_procstar_program.py::test_run_args_stringified
  (...)
  ======================================================================================================= 1 passed, 3 warnings in 0.03s =======================================================================================================
  ```

python/apsis/program/base.py
```diff
@@ -187,6 +195,13 @@ class RunningProgram:
 
     def __init__(self, run_id):
         self.run_id = run_id
+        self.args = {}
+
+    def set_run_args(self, args):
```
LudovicoRighi: nit/style: this `set_run_args` function doesn't do much; I'd suggest removing it and simply using `.args` where needed, especially if you end up end up removing `normalize_args` too (see [other comment](https://github.com/apsis-scheduler/apsis/pull/555/changes#r3978676900))

LudovicoRighi:
I'm adding some param name validation in this separate PR, as I think some special characters could break your env vars (i.e. `=` in param name). This is a pre-existing gap.
- https://github.com/apsis-scheduler/apsis/pull/565

Notice that currently nothing rejects the following:
```diff
> git diff
diff --git a/test/int/cond/jobs/many dep.yaml b/test/int/cond/jobs/many dep.yaml
index 3ba2dc6a..c54dd51e 100644
--- a/test/int/cond/jobs/many dep.yaml
+++ b/test/int/cond/jobs/many dep.yaml
@@ -1,4 +1,4 @@
-params: [val]
+params: [val=x]

 program:
   type: no-op
diff --git a/test/int/cond/test_dep.py b/test/int/cond/test_dep.py
index 0874667c..ef3642a5 100644
--- a/test/int/cond/test_dep.py
+++ b/test/int/cond/test_dep.py
@@ -93,7 +93,7 @@ def test_many_deps(inst, direction):
             {
                 "type": "dependency",
                 "job_id": "many dep",
-                "args": {"val": v},
+                "args": {"val=x": v},
             }
             for v in vals
         ],
@@ -116,7 +116,7 @@ def test_many_deps(inst, direction):

     # All dependencies but one.
     for v in vals[:-1]:
-        client.schedule("many dep", {"val": v})
+        client.schedule("many dep", {"val=x": v})

     time.sleep(1)
     # The run is still waiting.
@@ -124,7 +124,7 @@ def test_many_deps(inst, direction):
     assert res["state"] == "waiting"

     # Schedule the last dependency.
-    client.schedule("many dep", {"val": vals[-1]})
+    client.schedule("many dep", {"val=x": vals[-1]})

     # Now the run can continue.s
     res = inst.wait_run(run_id)
```
The tests would still pass.
```
[lrighi]🌐 ny3-rd025 in apsis [ lrighi/env-var-testing][$!?][📦 v2.4.3][🐍 v3.10.20][🅒 prod7][⏱ 2s]
> pytest test/int/cond/test_dep.py
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.10.20, pytest-8.4.1, pluggy-1.6.0
rootdir: /space/lrighi/asd-repos/apsis
configfile: pyproject.toml
plugins: anyio-4.9.0, asyncio-1.1.0
asyncio: mode=strict, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 11 items

test/int/cond/test_dep.py ...........                                                                                                                                                                                                 [100%]
(...)
====================================================================================================== 11 passed, 4 warnings in 15.64s ===============================================================================
```

LudovicoRighi:
There's some conflicts to solve. Have you pushed [7c0025f](https://github.com/apsis-scheduler/apsis/commit/7c0025f38ebff628ff2ec931046a4f0c079af328) by mistake? That change seems unrelated to this PR, so it shouldn't be included here.

(Also notice I've just merged https://github.com/apsis-scheduler/apsis/pull/551 which contains a similar fix)

LudovicoRighi (approved):

PR #557, opened 2026-09-02
Don't block the event loop while reloading jobs

fanman1:
This keeps reloading the jobs dir from freezing the scheduler.

First, it stops the reload from blocking the event loop. The old code pulled the whole directory listing into a list before awaiting anything, so the entire os.walk ran up front and froze the loop for several seconds on the NFS jobs dir. Now it walks the dir lazily and hands control back to the loop as it goes.

Second, it loads job files in batches of 16 instead of one at a time, so the reads and parses overlap. That makes the file loading part of the reload roughly 3x faster than doing one at a time.

Some rough numbers from the real asd/jobs repo with about 4700 files. These were measured on a local disk so the NFS numbers will be higher, and the box is shared so they bounce around.
- Max event loop blockage: the loop is blocked for max **~50ms** at a time, usually more like 15 to 25ms. That is how long it takes to walk one batch worth of directories between yields.
- Max wall clock: the whole reload lands around 5 to 11 seconds depending on box load

I chose a batch size of 16 because I ran over 500 time trials with batch sizes from 10 to 64, and they showed the wall clock was basically flat across the range but the blockage grew with batch size, so 16 gets almost all of the speedup while keeping the blockage small.

Addresses https://git.tudor.com/asd/iac/issues/1984. Bottom of a 2 PR stack, the libyaml parser speedup is #559 on top of this.

### DIFF
--- a/python/apsis/jobs.py
+++ b/python/apsis/jobs.py
@@ -13,6 +13,7 @@ from .actions import Action
 from .actions.schedule import successor_from_jso
 from .cond import Condition
 from .exc import JobError, JobsDirErrors, SchemaError
+from .lib import itr
 from .lib.json import to_array, to_narray, check_schema
 from .lib.py import tupleize, format_ctor
 from .program import Program, NoOpProgram
@@ -156,7 +157,9 @@ def dump_yaml(file, job):
 
 def list_yaml_files(dir_path):
     dir_path = Path(dir_path)
-    for dir, _, names in os.walk(dir_path):
+    for dir, dirs, names in os.walk(dir_path):
+        # Don't go into hidden dirs (e.g. `.git`)
+        dirs[:] = [d for d in dirs if not d.startswith(".")]
         dir = Path(dir)
         paths = (dir / n for n in names if not n.startswith("."))
         paths = (p for p in paths if p.suffix == ".yaml")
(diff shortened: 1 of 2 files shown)


python/apsis/jobs.py
```diff
@@ -270,10 +272,10 @@ def _parse():
             exc.job_id = job_id
             return job_id, None, exc
 
-    load_coros = [load_job(path, job_id) for path, job_id in list_yaml_files(jobs_path)]
-    for chunk in itr.chunks(load_coros, 100):
-        results = await asyncio.gather(*chunk)
-        for job_id, job, exc in results:
+    for chunk in itr.chunks(list_yaml_files(jobs_path), 16):
```
LudovicoRighi: nit: a constant for `16` would be better

python/apsis/jobs.py
```diff
@@ -157,7 +157,9 @@ def dump_yaml(file, job):
 
 def list_yaml_files(dir_path):
     dir_path = Path(dir_path)
-    for dir, _, names in os.walk(dir_path):
+    for dir, dirs, names in os.walk(dir_path):
```
LudovicoRighi: nit: I would suggest to name these variables `root, dirs, files`, which is more standard, i.e. matches the docs:
https://docs.python.org/3/library/os.html#:~:text=for%20root%2C%20dirs%2C%20files%20in%20os.walk(%27python/Lib/xml%27)%3A

LudovicoRighi (approved):
> These were measured on a local disk so the NFS numbers will be higher

We can easily move off of NFS; see lib/issues/19533.

gusostow (approved):

PR #559, opened 2026-09-03
Parse job files with a fast libyaml loader

fanman1:
Stacked on top of #557.

Replace ruamel's `YAML().load` with `DupCheckSafeLoader`, a libyaml (`yaml.CSafeLoader`) subclass, parsed in a thread. Preserves ruamel's behavior: duplicate-key rejection, YAML 1.2 scalar rules, merge keys, empty-scalar-as-null. Approx 24x faster parse and malformed YAML is now a per-job error instead of aborting the reload.

Supersedes #553 (closed when its branch was renamed). Prior review discussion is there.


fanman1:
Carrying over @gusostow's review question from #553

> Remind me, what's the performance payoff of using this custom loader? Tweaking it to this extent makes me a bit anxious. Remember, absolute job load performance isn't the top priority here. It's more not blocking the event loop and being 100% correct.

This PR contains only the new parser

- on the repo's 54 job files, the libyaml `DupCheckSafeLoader` parses a full reload in ~2.3ms vs ruamel's ~63ms, which is roughly **24–28x** faster. this doesn't affect blocking the event loop but it was a good speedup opportunity I noticed
- the loader is built to match ruamel exactly with duplicate-key rejection, YAML 1.2 scalar resolution, merge keys, and empty-scalar-as-null. I verified that every job file in the repo resolves identically under both loaders; the only file either rejects is `jobs/test/action/schedule.yaml` (an intentional duplicate-key fixture), and both reject it.

gusostow (approved):
I'm not really able to review this correctness but seems good after

1. Put it through a max effort AI review
2. All of our in-house job files parse identically

PR #562, opened 2026-09-04
Paginate the GET /runs endpoint to protect the event loop

fanman1:
## Problem

Queries over a large run history can block the event loop while the server reads and decodes the matching runs. The existing workaround rejects queries over 10,000 runs, but the reads below that limit still happen on the loop. This can delay scheduling.

See https://git.tudor.com/asd/iac/issues/2129.

## Approach

I paginated `GET /api/v1/runs` in the backend, where each request returns one page of runs plus a `paging.next` cursor for the next request. Runs are ordered newest first using their run number. That number is the immutable sqlite rowid, so pages descend by a stable key that never duplicates or skips a run, even as runs are written or move between the in-memory mirror and the database.

Each page is a keyset query, `job_id` with `rowid < cursor` ordered by `rowid` descending with a `LIMIT`, backed by a `(job_id, rowid)` index. The page size is fixed, so the read, decode, and response build all run on the loop, no worker thread or second connection. I benchmarked this on the largest production jobs: the index turns the newest-page query from a 50-430ms scan-and-sort into a ~1ms seek, and a bounded page decodes well within the loop's latency budget, so batching on the loop is enough and the offload isn't needed.

The index is created for new databases from the schema and backfilled on existing ones by `scripts/migrate-db.py`, rather than on open.

The client follows the cursor and collects all the pages before returning the consolidated list back to the CLI user, so the experience is the same but the loop is no longer blocked from reading every run at once. `Client.get_runs` also takes a `limit` to return just the newest N without walking the whole history, and `apsis runs` grows a `--limit` flag. `GET /jobs/<job_id>/runs` gets the same paged response and its client method walks the cursor too. Requests for explicit run IDs still return every requested run in one response, bounded by the number of IDs the caller lists. A malformed or oversized cursor is rejected with a 400.

## Tests

- `test/unit/test_api_paging_args.py` checks cursor parsing, including malformed, oversized, and repeated values.
- `test/unit/test_client_paging.py` checks the client following the cursor, stopping at `limit`, walking the job-runs endpoint, and failing loudly on a cursor that does not advance.
- `test/unit/test_run_db_query.py` checks database paging, ordering, filters, cursor exclusivity, and index creation for new and migrated databases.
- `test/unit/test_run_store.py` checks the memory and database merge without duplicates, transitions between pages, lookback filtering per page, and the bound on in-memory candidates.
- `test/int/pagination/test_pagination.py` drives a real server for the paging envelope, the client walking the pages, `--limit`, explicit run ID requests, and a 400 on a bad cursor.

### DIFF
--- a/python/apsis/runs.py
+++ b/python/apsis/runs.py
@@ -1,9 +1,11 @@
 from collections import namedtuple
+import copy
 import itertools
 import jinja2
 import logging
 import ora
 from ora import now, Time
+import re
 import shlex
 from typing import Iterator
 
@@ -347,6 +349,31 @@ def bind(run, job, jobs):
 # -------------------------------------------------------------------------------
 
 
+_RUN_ID_REGEX = re.compile(r"r([0-9]+)")
+
+
+def run_number(run_id):
+    """
+    Returns the integer N from a run_id of the form rN.
+
+    :raise ValueError:
+      `run_id` is not rN with N a non-negative integer.
+    """
+    match = _RUN_ID_REGEX.fullmatch(run_id) if isinstance(run_id, str) else None
+    if match is None:
+        raise ValueError(f"invalid run_id: {run_id!r}")
+    return int(match.group(1))
+
+
+def _snapshot(run):
+    """Returns a copy of `run` with its mutable dicts copied too."""
+    snap = copy.copy(run)
+    snap.times = dict(run.times)
+    snap.meta = dict(run.meta)
+    snap.run_state = copy.copy(run.run_state)
+    return snap
+
+
 class RunStore:
     """
     Storage API that stitches together cached in-memory runs and data from the DB. Additionally it keeps cached run
@@ -517,9 +544,28 @@ class RunStore:
           If True (default), applies lookback window. If False, queries all runs.
           Set to False for condition checks that need to see all active runs.
         """
-        # combine expected (scheduled, in-memory) and active (in-memory
-        # mirror of running/waiting/etc) into one filtered iterator of
-        # in-memory runs.
+        in_memory_list, db_kwargs = self.__prepare_query(
+            run_ids=run_ids,
+            job_id=job_id,
+            state=state,
+            since=since,
+            args=args,
+            with_args=with_args,
+            limit_lookback=limit_lookback,
+        )
+        in_memory_ids = {r.run_id for r in in_memory_list}
+
+        db_runs = [r for r in self.__query_run_db(**db_kwargs) if r.run_id not in in_memory_ids]
+
+        return now(), in_memory_list + db_runs
+
+    def __prepare_query(self, *, run_ids, job_id, state, since, args, with_args, limit_lookback):
+        """
+        Filters the in-memory runs and normalizes the DB filters.
+
+        :return:
+          `in_memory_list, db_kwargs` for `RunDB.query` or `RunDB.query_paged`.
+        """
         in_memory = itertools.chain(self.__expected_runs.values(), self.__active_runs.values())
 
         if state is not None:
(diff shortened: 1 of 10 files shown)


python/apsis/service/client.py
```diff
@@ -253,18 +268,18 @@ def mark(self, run_id, state_name):
         return self.__post("/api/v1/runs", run_id, "mark", state_name)
 
     def get_runs(self, *, job_id=None, state=None, args={}):
```
LudovicoRighi: Since `get_runs` is being touched here anyway: should we let callers pass `since=` and `limit=` through, and return a single page (no walk) when `limit` is given? That would let `apsis runs --job X` in `cli.py` stop pulling a job's entire history page by page, and would give the `# FIXME: add pagination` / `# FIXME: times` there (cli.py:221, 229) something to hook into. The CLI half could easily be a follow-up.

https://github.com/apsis-scheduler/apsis/blob/8f0702512e3252b68fa537cdcc5630e260960ea1/python/apsis/cli.py#L221

python/apsis/service/api.py
```diff
@@ -42,6 +42,61 @@
...
+    """
+    Pops query param `name` from `args`, allowing at most one value.
+
+    :return:
+      The single value, or None if absent.
+    :raise ValueError:
+      `name` was given more than once.
+    """
+    values = args.pop(name, None)
+    if not values:
```
gusostow: probably not a problem in this case but in general it's risky to depend on truthiness to detect None.

python/apsis/service/api.py
```diff
@@ -509,37 +570,25 @@ async def runs(request):
...
-    runs_list = list(runs)
-
-    def serialize_runs():
-        return runs_to_jso(request.app, when, runs_list, summary=summary)
+    def build():
+        runs, next_cursor = apsis.run_store.fetch_page(page_request, limit)
+        return runs_to_jso(request.app, when, runs, summary=summary), next_cursor
 
-    result = await asyncio.to_thread(serialize_runs)
+    result, next_cursor = await asyncio.to_thread(build)
```
gusostow: Noting the concurrent sqlite read transaction. It might be hard to repro with what I gave you since these queries will be quick. You might have better luck if you tweak it to be a query that will take a long time somehow.

python/apsis/runs.py
```diff
@@ -570,22 +616,80 @@ def query(
...
+            limit_lookback=True,
+        )
+        max_rowid = None if cursor is None else run_number(cursor)
+        if max_rowid is not None:
+            in_memory_list = [r for r in in_memory_list if run_number(r.run_id) < max_rowid]
+        # copy so the worker thread never sees a live run being mutated
+        in_memory_list = [_snapshot(r) for r in in_memory_list]
+        return self._PageRequest(in_memory_list, db_kwargs, max_rowid)
+
+    def fetch_page(self, request, limit):
```
gusostow: Apsis doesn't have much type hinting. Personally I don't think everything needs to be typed but I think we should include it where useful, particularly with structured objects that we have types for like _PageRequest.

I am eventually planning to assign you to do a first pass at adding useful type hints to apsis.

python/apsis/runs.py
```diff
@@ -570,22 +616,80 @@ def query(
...
+    def prepare_page(
+        self,
+        *,
+        run_ids=None,
+        job_id=None,
+        state=None,
+        since=None,
+        with_args=None,
+        cursor=None,
+    ):
```
gusostow: type hint return value

python/apsis/runs.py
```diff
@@ -570,22 +616,80 @@ def query(
...
+            run_ids=run_ids,
+            job_id=job_id,
+            state=state,
+            args=args,
+            with_args=with_args,
+            min_timestamp=min_ts,
+        )
+        return in_memory_list, db_kwargs
+
+    # split so the api runs the in-memory part on the loop and offloads the fetch to a thread
```
gusostow: I think these run store API changes are slightly tortured because of trying to make this split. Can you simplify if you end up not offloading the query to the background thread. If if we can't repro causing a deadlock I'm not sure if it's worth it for this extra code complexity.

python/apsis/sqlite.py
```diff
@@ -317,6 +298,11 @@ def canonical_args_json(args) -> str:
 # the raw JSON string is a direct index lookup.
 sa.Index("idx_runs_job_args", TBL_RUNS.c.job_id, TBL_RUNS.c.args)
 
+# index for keyset paging so a job's newest runs below a cursor are an index seek
+# not a scan and sort
+# rowid is the immutable run number so it is a stable cursor
+sa.Index("idx_runs_job_rowid", TBL_RUNS.c.job_id, TBL_RUNS.c.rowid)
```
gusostow, line 304: Have you benchmarked if this is necessary for our largest jobs? I suspect the scan is a minimal portion of the query time and most of it is deserialization which is fixed by doing this in small batches.

gusostow, line 304: This is how we've been handling migrations https://github.com/apsis-scheduler/apsis/blob/main/scripts/migrate-db.py.

gusostow (commented):
Looking good. Main thoughts are about if we should/need to be offloading the query if we're doing small batches. Think the PR can be a lot simpler if we give that up.

python/apsis/service/api.py
```diff
@@ -165,6 +220,9 @@ async def job(request, job_id):
 
 @API.route("/jobs/<job_id:path>/runs")
 async def job_runs(request, job_id):
+    # left unpaginated for now, its route is ambiguous with /jobs/<job_id:path> under
+    # sanic-routing and resolves nondeterministically
+    # deferred to a follow-up that also disambiguates the route, no current callers
```
LudovicoRighi: > no current callers

Notice `gold_status.py` in `asd/lib` (lines 602 and 707) polls this endpoint over raw HTTP with up to 8 workers, two calls per job per cycle, and `Client.get_job_runs()` still points here too. I opened #569 to handle the routing ambiguity, so should we give this route the same `prepare_page`/`fetch_page` treatment and `paging` envelope in this PR?

python/apsis/service/client.py
```diff
@@ -117,6 +117,21 @@ def __request(self, method, *path, timeout=None, data=None, **query):
...
+        Follows the `paging.next` cursor across pages, returning the merged
+        {run_id: run} dict.
+        """
+        runs = {}
+        cursor = None
+        while True:
+            # cursor is None on the first request, which __url drops
+            resp = self.__get(*path, cursor=cursor, **query)
+            runs.update(resp["runs"])
+            cursor = resp.get("paging", {}).get("next")
```
LudovicoRighi: Should the walk guard against a cursor that doesn't advance? If a server bug ever returned the same `next`, this loop would spin forever. One comparison with the previous cursor, raising on repeat, would make it fail loudly instead.

python/apsis/sqlite.py
```diff
@@ -869,14 +902,40 @@ def __init__(self, engine):
...
+        """
+        path = engine.url.database
+        if not path or path == ":memory:":
+            return engine
+
+        read_engine = sa.create_engine(
+            f"sqlite:///{path}",
+            poolclass=sa.pool.NullPool,
+            connect_args={"check_same_thread": False},
+        )
```
LudovicoRighi: **Preserve `database.timeout` on the reader connection**

The main engine receives the configured timeout, but this new engine only receives `check_same_thread=False`, so paginated reads use SQLite's default five seconds.

After `SqliteDB.open(path, timeout=0.125)`, `PRAGMA busy_timeout` reports 125 ms on the primary connection and 5,000 ms on the reader. Configurations above five seconds are also shortened. Reads that encounter a SQLite lock therefore wait or fail on a different schedule from the configured one.

Suggestion: pass the normalized timeout into the read-engine constructor too, and verify the setting on both connections. Ordinary WAL reads coexist with scheduler writes; this is a configuration regression, not evidence of a general reader/writer deadlock.

python/apsis/runs.py
```diff
@@ -570,22 +616,80 @@ def query(
...
+            since=since,
+            args=None,
+            with_args=with_args,
+            limit_lookback=True,
+        )
+        max_rowid = None if cursor is None else run_number(cursor)
+        if max_rowid is not None:
+            in_memory_list = [r for r in in_memory_list if run_number(r.run_id) < max_rowid]
+        # copy so the worker thread never sees a live run being mutated
+        in_memory_list = [_snapshot(r) for r in in_memory_list]
```
LudovicoRighi: **Bound the in-memory candidates before snapshotting them**

`prepare_page` does not receive `limit`, so this copies every matching in-memory run below the cursor on the event loop. Only the later worker-side merge truncates the results. A 500-result request can therefore copy tens of thousands of runs synchronously.

Suggestion: pass the limit into preparation, select the newest `limit + 1` matching in-memory candidates below the cursor, and snapshot only those before crossing into the worker. That is sufficient alongside the database lookahead for the merge. Keep live mutable runs on their owning thread. This bounds the expensive copying; avoiding repeated full-cache scans would additionally require an ordered lookup structure.

A useful regression check would populate many matching expected/active runs, request a small page, and assert that at most `limit + 1` candidates are snapshotted while retaining the existing merge checks.

python/apsis/sqlite.py
fanman1: yup this is still necessary for significantly cutting down on the scan time down to ~1 ms instead of 50-400ms

LudovicoRighi:
@fanman1 this commit https://github.com/apsis-scheduler/apsis/pull/562/changes/83ce280a8667f7805a10261682084879fed7fc7c contains lots of unrelated changes and it's also quite big; it's good practice to have one commit per change, especially after the reviewers have already done one pass on the PR. It makes it much easier for them to do another pass of review.

python/apsis/service/client.py
```diff
@@ -117,6 +117,33 @@ def __request(self, method, *path, timeout=None, data=None, **query):
...
+        while True:
+            # cursor is None on the first request, which __url drops
+            resp = self.__get(*path, cursor=cursor, **query)
+            runs.update(resp["runs"])
+            if max_runs is not None and len(runs) >= max_runs:
+                return dict(list(runs.items())[:max_runs])
+            next_cursor = resp.get("paging", {}).get("next")
+            if next_cursor is None:
+                return runs
+            if next_cursor == cursor:
```
gusostow: Should this reject cursors that increase also?

gusostow (approved):

PR #566, opened 2026-09-10
Validate job param names

LudovicoRighi:
A job param must now be a valid identifier (not a keyword) and must not be one of the names Apsis itself provides to templates (run_id, job_id, and the BIND_ARGS helpers).  Any other name can't be referenced in a template expansion, and names containing "=" can't be written as NAME=VAL in the CLI or the web UI.  A job with an invalid param name fails to load with a SchemaError, so it's caught by check-jobs and at reload.

Run args are not validated separately: validate_args already rejects any arg that isn't a param.  Runs stored by earlier versions may carry arg names that are no longer valid; they load, restore, and are served unchanged.  A stored ad hoc job with such a param now fails to load from the DB and is reported as an unknown job (consistent with JobDB.query) instead of raising.

The `apsis schedule` CLI reports a clean error for an arg without "=".


LudovicoRighi:
Closing in favor of PR 565.

PR #568, opened 2026-09-11
record the run level reason on error and failure transitions

fanman1:
When a run enters the error or failed state, Apsis already knows why, but the reason mostly stays in the run log. The run log isn't available to the ops incident action, which runs on a separate thread and can't reach Apsis, so incidents couldn't show why a run failed.

Every transition to error or failure now records a human-readable reason in `run.meta["state_reason"]`, which already flows into the run snapshot the incident action reads. It's set from the program result (exit code, killed by signal, runtime timeout), agent connect and reconnect failures, dependency timeouts and missing dependencies, restore recovery, manual marks, and the generic exception branch. A later transition clears it, so a rerun to success doesn't keep a stale reason. Program-level diagnostics still live in `program.errors` (Procstar's own error list) and the reason here is not accumulated into that list.

test/unit/test_running.py covers the reason for program failures, program errors, and internal errors, and that a later transition clears it. test/int/procstar/test_program.py checks that a reconnect failure surfaces the reason in `state_reason` while keeping the run's program metadata intact.

for https://git.tudor.com/asd/iac/issues/2182 and https://git.tudor.com/asd/lib/issues/18395. the lib formatter reads `meta.state_reason` to lead the ops incident in https://git.tudor.com/asd/lib/pull/24957. deploy apsis first so the field exists.

### DIFF
--- a/python/apsis/running.py
+++ b/python/apsis/running.py
@@ -46,28 +46,20 @@ async def _maybe_compress(outputs, *, compression="br", min_size=16384):
     return dict(zip(outputs.keys(), o))
 
 
-def _program_meta(meta):
+def _program_meta(meta, *, error_message=None):
     """
     Returns a run metadata update for program metadata `meta`.
 
     A run's program metadata is replaced by, rather than merged with, the
-    metadata of each program update, so a program that reports no metadata must
-    produce no update at all.  Otherwise it would erase what is already known
-    about the program, such as the host and pid of a process that Apsis has
-    lost track of.
+    metadata of each program update, so an update with no metadata or error
+    message must produce no update at all.  Otherwise it would erase what is
+    already known about the program, such as the host and pid of a process that
+    Apsis has lost track of.
     """
+    if error_message is not None:
+        meta = meta or {}
+        meta = {**meta, "errors": [*(meta.get("errors") or []), str(error_message)]}
     return {"program": meta} if meta else {}
-
-
-def _program_error_meta(run, error):
-    """Include the error message while preserving available program metadata."""
-    meta = error.meta or run.meta.get("program", {})
-    return {
-        "program": {
-            **meta,
-            "errors": [*(meta.get("errors") or []), str(error.message)],
-        },
-    }
 
 
 async def _process_updates(apsis, run):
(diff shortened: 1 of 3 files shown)


python/apsis/apsis.py
```diff
@@ -376,17 +383,27 @@ def _update_output_data(self, run, outputs, persist):
...
+          Human-readable reason for the transition, recorded as
+          meta["state_message"] so snapshots and actions can show it.  A
+          later transition with no message clears it.
         """
         time = now()
         run_id = run.run_id
 
+        # stash the reason so snapshots and actions can show it
+        if message is not None:
+            meta = {**meta, "state_message": str(message)}
```
gusostow: nit: can you rename this to `reason` and `state_reason` to be a little more specific?

python/apsis/running.py
```diff
@@ -46,16 +46,19 @@ async def _cmpr(output):
...
-    about the program, such as the host and pid of a process that Apsis has
-    lost track of.
+    metadata of each program update, so an update with no metadata or error
+    message must produce no update at all.  Otherwise it would erase what is
+    already known about the program, such as the host and pid of a process that
+    Apsis has lost track of.
     """
+    if error_message is not None:
+        meta = meta or {}
+        meta = {**meta, "errors": [*(meta.get("errors") or []), str(error_message)]}
```
gusostow: Do we actually need to accumulate this list of errors if we're dropping it in the state_reason?

This code is a little hard to reason about so wondering if we can just drop the this errors change.

gusostow: I think I'm confused about what this is doing. I believe it's making sure that previous errors aren't dropped if procstar reports multiple errors? Can you give an example of what this does for us

test/unit/test_running.py
```diff
@@ -163,18 +164,21 @@ async def test_error_keeps_startup_metadata():
...
-            ProgramError("start failed: proc0: no open connection in group"),
+            error,
         ]
     )
 
     await _process_updates(apsis, run)
 
     assert run.state == State.error
     assert run.meta["program"] == STARTUP_META
+    # the reason is carried in state_reason, not appended to program metadata
```
gusostow: I think these comments are somewhat path dependent within this PR. Future readers won't know that you were  briefly trying out appending the errors to the program metadata

gusostow (approved):

PR #569, opened 2026-09-15
Fix job run routing and reject /runs suffixes in check-jobs

LudovicoRighi:
`GET /api/v1/jobs/<job_id>/runs` could return either the job handler's response
or run history, depending on Sanic's arbitrary registration order for two
overlapping path routes. Register one route and dispatch on the raw literal
`/runs` suffix so requests consistently reach the run-history handler.

Percent-encoded suffixes remain job ID data: `/jobs/X%2Fruns` fetches the job
`X/runs`, and `/jobs/X%2Fruns/runs` fetches its runs. Encode job IDs in both Python
client methods so `apsis job X/runs` returns that job's definition. Document this
contract and cover it with raw HTTP, client, and CLI tests using distinct jobs
and run IDs.

`apsisctl check-jobs` also rejects IDs ending exactly in `/runs`, reports every
offending job, and exits with status 1. Top-level `runs`, `dryruns`,
`group/test-runs`, and `group/runs/step` remain allowed. This check is confined
to the command: startup and reload still accept existing jobs, and encoded
IDs remain supported by the API and client.

The regression suite asserts that no overlapping jobs route is registered;
unlike sampling fresh interpreters, this reliably detects the old route layout.
The test was verified to fail against the PR base. The client tests also
reproduced the previous silent CLI failure before the encoding fix.

Validation: 42 relevant integration tests passed locally, covering the new
command checks, existing job validation, routes, client, and CLI. The four new
cases verify rejection with and without dependency checking, allowed names,
and loader compatibility. Python formatting passes. Ruff reports no new
diagnostics; `ctl.py` retains its pre-existing F401 for the required Sanic
import. Full CI still needs to run on the updated PR head.


python/apsis/service/api.py
```diff
@@ -152,8 +152,23 @@ def match_job_id(jobs, job_id):
...
 
 
+# Job IDs may contain slashes, so `job_id` is a `path` param and it captures
+# a trailing `/runs` too.  `/jobs/<job_id:path>` and `/jobs/<job_id:path>/runs`
+# are therefore ambiguous under sanic-routing, and since sanic 21.6 keeps a
+# blueprint's routes in a set, the two were registered in arbitrary order and
+# `GET /jobs/X/runs` resolved to either handler depending on the process.  A
+# single route with an explicit dispatch on the suffix makes the choice
+# independent of registration order.  The dispatch looks at the raw path, so a
+# percent-encoded `%2Fruns` is still part of a job ID; a job whose ID itself ends
```
fanman1: add a test for the %2Fruns case?

fanman1 (approved):

fanman1 (approved):
lgtm!

python/apsis/check.py
```diff
@@ -59,13 +59,18 @@ def get(self, job_id):
         raise LookupError(job_id)
 
 
-def check_job(jobs_dir, job):
+def check_job(jobs_dir, job, *, check_job_names=False):
```
gusostow: Any reason to not just always do this?

gusostow (approved):

python/apsis/check.py
LudovicoRighi: Yeah better

PR #570, opened 2026-09-18
gate ad hoc jobs

fanman1:
## Problem

Ad hoc jobs let callers submit commands directly through the CLI or API, which can bypass review of registered job definitions as described in [the issue](https://git.tudor.com/asd/iac/issues/2203). We still want that convenience for local testing, so removing the feature entirely would be a poor fit.

## Approach

Add `adhoc.enabled` with a default of `false` and require a boolean value, while keeping ad hoc jobs available when an environment explicitly enables them. Disabled servers reject submitted job definitions with HTTP 403 before loading or saving them, and stored ad hoc jobs enter the error state if scheduled again, rerun, or started from the queue.

Queued runs retain their ad hoc status so registered jobs still start if their definitions are removed, with a fallback for runs saved by older versions. Already running jobs can finish, and the documented setting takes effect after restarting Apsis.

## Test

- `test/int/test_adhoc.py` checks CLI and API denials, stored and queued runs, older saved runs, and registered jobs starting after their definitions are removed.
- `test/int/test_config.py` checks the default, explicit boolean values, and invalid settings.
- `test/int/test_transitions.py` checks existing run transitions with ad hoc jobs enabled.
- `test/int/procstar/test_sudo.py` checks that enabled ad hoc jobs still support sudo programs.


docs/config.rst
```diff
@@ -73,6 +76,23 @@ are not held in memory and are not visible in user interfaces.  They are
...
+``apsis adhoc`` or submit job definitions through the API, for example in a
+local development environment.
+
+When disabled, API requests containing a job definition return HTTP 403 before
+the job is loaded or saved.  Scheduling or rerunning a stored ad hoc job creates
+a run in the error state, and queued ad hoc runs transition to error when they
+reach their start time or are manually started.  Already running jobs can
+finish, and registered jobs are unaffected.
+
+Restart Apsis after changing this setting.
```
gusostow: This is assumed for every setting

python/apsis/apsis.py
```diff
@@ -242,6 +242,17 @@ def _start(self, run):
         Runs the run's program in a task added to `__run_tasks`.
         """
         assert run.run_id not in self._running_programs
+        if not self.cfg.get("adhoc", {}).get("enabled", False):
```
gusostow: My take is that it's overkill to fail rerunning previous adhoc jobs. Don't think it'll save us anything and every line of code has a cost

gusostow (approved):

